PR #1, opened 2017-08-03
HTTP server & Abstract Command Handler

han-gyeol:
WIP don't merge!


.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
+{
```
saghul: Please drop this. Add .vscode to .gitignore instead.

css/animate.css
```diff
@@ -0,0 +1,1579 @@
+@charset "UTF-8";
```
saghul: Don't bundle any CSS libraries. If we need one, we must use an npm package for it.

main.js
```diff
@@ -0,0 +1,68 @@
...
+
+/**
+ * Options used when creating the main Jitsi Meet window.
+ */
+const jitsiMeetWindowOptions = {
+    width: 800,
+    height: 600,
+    minWidth: 800,
+    minHeight: 600,
+    titleBarStyle: 'hidden'
```
saghul: the app will be used in full-screen mode, so let's skip this for now.

main.js
```diff
@@ -0,0 +1,68 @@
+const electron = require('electron');
+const APP = electron.app;
+const BrowserWindow = electron.BrowserWindow;
+
+const path = require('path');
```
saghul: style: put imports first, global constants after

windows/jitsi-meet/index.html
```diff
@@ -0,0 +1,53 @@
...
+<html>
+    <head>
+        <style>
+            body, html {
+                margin: 0;
+                height: 100%;
+                overflow: hidden;
+            }
+            body {
+                -webkit-app-region: drag
```
saghul: we won't need this if the titlebar is not hidden. in addition, doing this breaks buttons a bit.

windows/jitsi-meet/index.html
```diff
@@ -0,0 +1,53 @@
...
+        <link rel="stylesheet" href="../../css/animate.css">
+        <script src="https://meet.jit.si/external_api.js"></script>
+        <script>
+            require("./render.js");
+        </script> 
+    </head>
+    <body>
+        <div id="background_pattern"></div>
+        <div id="background_logo"></div>
+        <div id="cs-loader">
```
saghul: Let's cleanup everything here. Everything is going to be replaced by the calendar view, so just show the background a logo.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+function createJitsiIframe(room) {
+    const domain = 'meet.jit.si';
+    const options = {
+        roomName: room,
+        width: 800,
+        height: 600,
+        parentNode: document.querySelector('#meet')
+    };
+
+    /* eslint-disable */
```
saghul: you can add this to the `/* global: ` section instead.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * Cteates the iframe that will load Jitsi Meet.
+ *
+ * @param {string} room - conference room name
+ * @returns {null}
+ */
+function createJitsiIframe(room) {
+    const domain = 'meet.jit.si';
```
saghul: let's take this from some config.js module or an env variable. Also when we require the script.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
+const server = require('../../modules/httpcontrol').HttpServer;
+const listeningPort = 1025;
```
saghul: take this from an env variable in the HTTP server itself.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+ */
+let api;
+
+window.onload = function() {
+    server.init(listeningPort);
+    server.onReceivedCommand((type, args) => {
+        handleCommand(type, args);
+    });
+
+    document.onkeydown = _onKeyDown;
```
saghul: please remove this, commands should only be sent by the HTTP server.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+        );
+    };
+
+    /**
+     * Listener for Eneter key down event.
+     *
+     * @param {event} event - key down event
+     * @returns {null}
+     */
+    function _onKeyDown(event) {
```
saghul: ditto

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+};
+
+/**
+ * Loads Jitsi-meet page in iframe with given room name.
+ *
+ * @returns {null}
+ */
+function triggerLoader() {
```
saghul: drop this too.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,167 @@
...
+
+/**
+ * Sends join jitsi-meet conference http request
+ *
+ * @param {string} url - target server url
+ * @param {string} command - command for JitsiMeetExternalAPI
+ * @param {Function} callback - callback function
+ * @returns {null}
+ */
+function sendHttpCommand(url, command, callback) {
```
saghul: this won't be needed then

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,125 @@
...
+     * Initializes the http server.
+     * 
+     * @param {integer} port - listening port for http server
+     * @returns {null}
+     */
+    init(port) {
+        const self = this;
+
+        this.started = true;
+        this.server = require('http').createServer((req, res) => {
```
saghul: put the require on top

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,125 @@
...
+    /**
+     * Initializes the http server.
+     * 
+     * @param {integer} port - listening port for http server
+     * @returns {null}
+     */
+    init(port) {
+        const self = this;
+
+        this.started = true;
```
saghul: you are doing this in start

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Initializes the http server.
+     * 
+     * @param {integer} port - listening port for http server
+     * @returns {null}
+     */
+    init(port) {
+        const self = this;
```
saghul: use an arrow function to avoid this

windows/jitsi-meet/index.html
```diff
@@ -0,0 +1,15 @@
+ <!DOCTYPE html>
+<html>
+    <head>
+        <link rel="stylesheet" type="text/css" href="../../css/welcome.css">
+        <script src="https://meet.jit.si/external_api.js"></script>
```
saghul: this needs to be based on the domain. Instead of loading it here, lazy load it with https://github.com/feross/load-script2

windows/jitsi-meet/index.html
```diff
@@ -0,0 +1,15 @@
...
+        <script>
+            require("./render.js");
```
saghul: lets call that file main.js

windows/jitsi-meet/index.html
```diff
@@ -0,0 +1,15 @@
...
+        </script> 
+    </head>
+    <body>
```
saghul: The rest of the UI components (like the calendar) are going to be React based. Let's start using React right here right now! :-)

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+/**
+ * Handles JitsiMeetExternalAPI command.
+ *
+ * @param {string} type - command type
+ * @param {string} args - arguments for the command
+ * @returns {string} execution status message
+ */
+function handleCommand(type, args) {
```
saghul: move the entire commands handling logic to a separate file please.

windows/jitsi-meet/render.js
```diff
@@ -0,0 +1,94 @@
...
+ */
+function createJitsiIframe(room) {
+    const options = {
+        roomName: room,
+        width: 800,
+        height: 600,
+        parentNode: document.querySelector('#meet')
+    };
+
+    api = new JitsiMeetExternalAPI(config.JITSI_MEET_DOMAIN, options);
```
saghul: you need to add listener to know when the conference was left so you can dispose the api object.

windows/jitsi-meet/index.html
han-gyeol: But there is already main.js file exist in the root directory. I put render.js as the name because it runs in the render process of Electron. Maybe change name to app.js instead?

saghul: Yeah, that sounds about right!

saghul:
@han-gyeol GitHub doesn't send notifications when you push new commits. Let me know when you want me to take another look.

windows/jitsi-meet/index.html
han-gyeol: Will it be better if I use this? https://github.com/saghul/jitsi-meet-api

windows/jitsi-meet/render.js
han-gyeol: Actually I don't dispose the api when the user hangup the conference. It just goes back to the Jitsi Meet welcome page. Should I dispose api when the App is closed instead?

modules/commandhandler/index.js
```diff
@@ -0,0 +1,81 @@
+const config = require('../../config.js');
+const load = require('load-script2');
+
+/**
+ * The jitsi-meet api.
+ */
+let api;
```
han-gyeol: I moved the api reference to the command handler module because I don't really find a reason it needs to be in app.js file. Will this be alright?

.eslintrc.js
```diff
@@ -5,23 +5,28 @@ module.exports = {
...
+        'eslint:recommended', "plugin:node/recommended", 'plugin:react/recommended'
     ],
     'globals': {
         // The globals that (1) are accessed but not defined within many of our
         // files, (2) are certainly defined, and (3) we would like to use
         // without explicitly specifying them (using a comment) inside of our
         // files.
-        '__filename': false
+        '__filename': false,
+        'JitsiMeetExternalAPI': true
```
hristoterezov: Usually we are defining this type of globals in the files where they are used with comment in the beginning of the file:
`/* global JitsiMeetExternalAPI */`

.jshintrc
```diff
@@ -0,0 +1,10 @@
+{
```
hristoterezov: Could you please remove the jshint? Why do we need it?

main.js
```diff
@@ -0,0 +1,68 @@
...
+/**
+ * Options used when creating the main Jitsi Meet window.
+ */
+const jitsiMeetWindowOptions = {
+    width: 800,
+    height: 600,
+    minWidth: 800,
+    minHeight: 600,
+    titleBarStyle: 'hidden',
+    webPreferences: { experimentalFeatures: true }
```
hristoterezov: Why do you need `experimentalFeatures: true`?

windows/jitsi-meet/app.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import ReactDom from 'react-dom';
+
+import http from '../../modules/httpcontrol';
+import handleCommand from '../../modules/commandhandler';
+import config from '../../config.js';
+import App from '../../components';
+
+const listeningPort = config.PORT;
```
hristoterezov: I'd rename the constant to `LISTENING_PORT`. It looks clearer to me that this is a constant.

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,123 @@
+import { EventEmitter } from 'events';
+import http from 'http';
+import querystring from 'querystring';
+
+/**
+ * Receives http requests from another peer.
+ */
+class HttpControl extends EventEmitter {
```
hristoterezov: Do we really need so much logs? Even if we need it maybe we should use lower level of logging for some of the messages. WDYT?

modules/commandhandler/index.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Creates the iframe that will load Jitsi Meet.
+ *
+ * @param {string} room - conference room name
+ * @returns {void}
+ */
+function createJitsiIframe(room) {
+    load(config.EXTERNAL_API_DOMAIN, err => {
+        if (err) {
+            console.log(`Error creating Jtisi-Meet iframe: ${err}`);
```
hristoterezov: Isn't it better to return the error? Currently if there is an error `handleCommand` will still return 'success....'.

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,123 @@
...
+        this.started = false;
+    }
+
+    /**
+     * Event listener for join conference request.
+     * 
+     * @param {Function} callback - callback function
+     * @returns {void}
+     */
+    onReceivedCommand(callback) {
```
hristoterezov: Why do we need this method? Can't you just use the event 'command'?

.jshintrc
han-gyeol: My mistake. Will remove it soon

main.js
han-gyeol: I used it to test a feature but forgot to remove. Will remove it also

modules/httpcontrol/index.js
han-gyeol: I put logs for the development purpose, but I guess some of them can be removed now. What sort of lower-level logging would you suggest?

modules/commandhandler/index.js
han-gyeol: You're right. I will make it return error instead

modules/httpcontrol/index.js
```diff
@@ -0,0 +1,123 @@
...
+            console.log('Server listening at port %d', port);
+        });
+    }
+
+    /**
+     * Starts processing the events.
+     * 
+     * @returns {void}
+     */
+    start() {
```
han-gyeol: I put it in case there is a need to disable http temporarily, and use other protocols instead. There is no immediate need yet. Should I remove it?

modules/httpcontrol/index.js
han-gyeol: There is no practical reason except than I get to define the jsdoc description for that method. The method will work fine with just on('command') as well

han-gyeol:
I wonder if I should redux to fire up the JitsiMeetExternalAPI iframe? Other than that, I don't see any UI components that need to be mapped to the state.

modules/commandhandler/index.js
```diff
@@ -0,0 +1,103 @@
...
+ * @returns {void}
+ */
+function createJitsiIframe(room) {
+    load(config.EXTERNAL_API_DOMAIN, err => {
+        if (err) {
+            return err;
+        }
+        const options = {
+            roomName: room,
+            width: 800,
```
saghul: set these to '100%';

modules/commandhandler/index.js
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * Creates the iframe that will load Jitsi Meet.
+ *
+ * @param {string} room - conference room name
+ * @returns {void}
+ */
+function createJitsiIframe(room) {
+    load(config.EXTERNAL_API_DOMAIN, err => {
+        if (err) {
+            return err;
```
saghul: we should show an alert on the screen when this happens

modules/commandhandler/index.js
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Handles JitsiMeetExternalAPI command.
+ *
+ * @param {string} type - command type
+ * @param {string} args - arguments for the command
+ * @returns {string} execution status message
+ */
+function handleCommand(type, args) {
+    if (api) {
```
saghul: this code branches too much and it's hard to read. can you try to think of a way to make it simpler?

modules/httpcontrol/index.js
saghul: we would use the jitsi-meet-logger, which has configurable levels

hristoterezov: Maybe info or debug ...

modules/httpcontrol/index.js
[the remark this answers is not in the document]
hristoterezov: I would remove them and put the start logic into the constructor. You can also do a dispose method if there is something that need to be cleaned up. This will cause exporting the class and creating the object where you are calling start now.

PR #4, opened 2018-01-24
fix(remote-control): toggle screenshare button state on click

virtuacoplenny:


src/features/remote-control-menu/remote-control-menu.js
```diff
@@ -20,6 +20,11 @@ export default class RemoteControlMenu extends React.Component {
     constructor(props) {
         super(props);
 
+        // FIXME: temporarily store this in state
```
bbaldino: i think it'd be good to add more details to this FIXME (what SHOULD be done, what must happen before it could be done that way, etc.)

PR #6, opened 2018-01-25
feat(ultrasound): use sound to navigate to remote control page

virtuacoplenny:
- Transmitter loads on the calendar page at emits the remote
  control url at an interval. The remote control listener page
  loads the receiver and waits for the transmitter's message.
- Quiet.js seems to have some assumptions about its file locations
  that I have not fully groked so they are at the top level for
  now. These assumptions were preventing it from being loaded
  through the traditional webpack bundling.
- Config was opened up to share constants between different
  files. It was needed for Quiet.js initialization. Before it
  was in gitignore to prevent my test client_id from being
  exposed but I realized it's not really that private and is
  a potential throwaway test client anyways.


README.md
```diff
@@ -6,28 +6,20 @@ A web frontend for integrating a google calendar with jitsi meetings. The intent
...
 
 ### Installing
 
 To build the frontend, run `npm install` to download dependencies and `npm run build-prod` to build the javascript bundle. Next serve `index.html` from a server, such as with `python -m SimpleHTTPServer`.
 
 If developing, running `npm run watch-dev` to build the javascript will run webpack in watch mode to automatically rebundle on changes and also enable debug features.
 
+### Special note about ultrasound support
+
+The app currently uses Quiet.js for ultrasound. The spot client can emit an ultra sound message for remote clients to be notified of which remote control url to go to. The Quiet.js files are currently in the top level of the project and should not be modified directly as they are third-party files.
```
bbaldino: `remote control url` is a little confusing i think.  is it right that this is referring to "the url you go to on any web client to remotely control the spot in order to join a call, etc"?

quiet.js
```diff
@@ -0,0 +1,947 @@
+/* Copyright 2016, Brian Armstrong
+ * quiet.js includes compiled portions from other sources
```
bbaldino: why does this get committed and not pulled in as a dependency?

src/features/ultrasound/receiver.js
```diff
@@ -0,0 +1,147 @@
...
+    _resetCachedContent() {
+        this._rawContent = new ArrayBuffer(0);
+        this._textContent = '';
+    }
+
+    /* eslint-disable max-len */
+    // An example valid format is:
+    // ABCDEFGH-ABCD-ABCD-ABCD-ABCDEFGHIJKL@lenny.jitsi.net/OPQRSTUV-OPQR-OPQR-OPQR-OPQRSTUVWXYZ
+    /* eslint-enable max-len */
+    _isValidJid(jid) {
```
bbaldino: strophe has nothing for this?

quiet.js
virtuacoplenny: quiet.js isn't set up for bundling and, at the time, I didn't feel like debugging why quiet.js was complaining when I had all the files nested in a thirdparty-libs directory. I'm gonna look into other ultrasound libraries now, plus I have to demo this prototype soon, so stuffed them into the root for now. I'll re-examine the ultrasound library placement if quiet.js is kept and this pull request is desired to be merged.

src/features/ultrasound/receiver.js
virtuacoplenny: I didn't see it in the docs and I asked for advice and was told to try length counting as a start. I did see strophe has some utils for parsing out domain/barejid/etc so I'll try to use those helpers.

virtuacoplenny:
Ultrasound only worked when the listener was put right next to the speaker. There is no pressure to get this feature in so I'm closing this PR.

PR #9, opened 2018-11-19
fix(meeting): allow configuring multiple meeting domains

virtuacoplenny:
The app before was configured to always go to one hardcoded
domain. Now allow a list of domains.

### DIFF
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -31,16 +31,9 @@ export function parseMeetingUrl(url) {
     const { host, pathname } = new URL(url);
     const pathParts = pathname.split('/');
     const roomName = pathParts.pop();
-    const newPath = pathParts.reduce((accumulator, currentValue) => {
-        if (currentValue) {
-            return `${accumulator}/${currentValue}`;
-        }
-
-        return accumulator;
-    }, '');
 
     return {
-        path: newPath,
+        path: pathParts.join('/'),
         host,
         roomName
     };
--- a/src/utils/url.js
+++ b/src/utils/url.js
@@ -1,0 +1,42 @@
+/**
+ * Checks if a passed in url can be opened in the application.
+ *
+ * @param {string} meetingUrl
+ * @param {Array<string>} validHosts
+ * @return {boolean}
+ */
+export function isValidMeetingUrl(meetingUrl, validHosts) {
+    try {
+        const { domain, host, roomName } = parseMeetingUrl(meetingUrl);
+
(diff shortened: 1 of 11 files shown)


src/utils/url.js
```diff
@@ -0,0 +1,42 @@
...
+        const { domain, host, roomName } = parseMeetingUrl(meetingUrl);
+
+        return domain && roomName && validHosts.includes(host);
+    } catch (e) {
+        return false;
+    }
+}
+
+/**
+ * Parses a meeting url so its parts can be accessed through meeting-related
```
bbaldino: can you give an example of a url arg and then how it is split into domain/host/roomname

src/views/meeting.js
```diff
@@ -31,24 +34,61 @@ export class Meeting extends React.Component {
...
+     */
+    _getMeetingUrl() {
+        const queryParams = new URLSearchParams(this.props.location.search);
+        const meetingUrl = queryParams.get('location');
+
+        if (!meetingUrl) {
+            return null;
+        }
+
+        if (isValidMeetingUrl(meetingUrl, VALID_MEETING_HOSTS)) {
```
bbaldino: should `isValidMeetingUrl` just handle null as well?

virtuacoplenny: It returns false for null, if that answers your question. The explicit return of null above is to handle an error case where somehow the user landed on the /meeting route without a meeting defined (a case that so far can only be reproduced by manually going to the /meeting url).

src/views/meeting.js
```diff
@@ -31,24 +34,65 @@ export class Meeting extends React.Component {
...
+    /**
+     * Parses the current window location for a valid meeting url. The value
+     * {@code null} will be returned if no valid meeting url is found.
+     *
+     * @private
+     * @returns {string|null}
+     */
+    _getMeetingUrl() {
+        const queryParams = new URLSearchParams(this.props.location.search);
+        const meetingUrl = queryParams.get('location');
```
bbaldino: after reading the stuff below more closely, this reads a little funny to me.  because we check `isValidMeetingUrl` and if that fails, it seems like it'd be an error but really we just treat it as a meeting name which isn't clear.  i'm thinking we do something like:

```
const location = queryParams.get('location');

if (isValidMeetingUrl(location, VALID_MEETING_HOSTS)) {
  return location;
} else if (isValidMeetingName(location)) {
  return `https://${VALID_MEETING_HOSTS[0]}/${meetingUrl}`;
}

return null;
```

PR #12, opened 2018-11-20
ref(config): access through redux only

virtuacoplenny:
Right now the config exists in a file built into the repo and overridable by environment variables at build time. This change makes it so accessing of config values, aside from initial redux hydration, is done through redux to avoid direct access.


virtuacoplenny:
It'll be easier for me to just redo this commit than fix the conflicts.

PR #15, opened 2018-11-20
ref(loading): higher-order component for loading

virtuacoplenny:


virtuacoplenny:
Force pushed to fixup all PR feedback and to fix the following conflicts:
- src/app.js: ErrorBoundary addition conflicted with imports and route declarations
- src/views/index.js: ErrorBoundary exporting conflicted with removal of Loading view
- src/views/loading.js: No longer passing dispatch into remote control service conflicted with the Loading view being removed

PR #17, opened 2018-11-21
fix(lint): add process to global in webdriver

virtuacoplenny:


virtuacoplenny:
Jenkins triggering is working now. I think what's left for now is giving jenkins permission to write to the spot repo so the result can be seen in the PR.

damencho:
I just updated the repo and jenkins user should be able to write.

PR #18, opened 2018-11-21
feat(tests): script to run all tests on ci

virtuacoplenny:


virtuacoplenny:
There are some quirks with launching chrome on the jenkins machine. For now the jenkins job will manually call install/lint/test. I'm thinking when ready jenkins can start executing the script with environment variables passed in. Until then, the script works for me locally so I'm thinking of committing it. @bbaldino, thoughts?

scripts/ci.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+run('npm install');
+run('npm run lint');
+run('npm test');
+run('npm run build-prod');
+
+console.log('Starting test server');
+
+const testServer = childProcess.spawn('python', [
```
bbaldino: it looks like this would be handy to have as a target in package.json, and it could take the port as an argument

virtuacoplenny:
Force pushed to resolve conflicts in webpack.config.js after webpack got upgraded from 3 to 4. Also fixed up converting the script from node/js to shell/bash.

PR #20, opened 2018-11-24
feat(tests): acceptance test for entering a meeting name and joining

virtuacoplenny:
This PR is built on top of PR #15, which should go in first and then I'll deal with the conflicts. This PR is for the new webdriver tests for manually entering a meeting name and joining the meeting (last commit).

### DIFF
--- a/ui-tests/page-objects/meeting-input.js
+++ b/ui-tests/page-objects/meeting-input.js
@@ -1,0 +1,54 @@
+const PageObject = require('./page-object');
+
+const INPUT = '[data-qa-id=meeting-name-input]';
+const SUBMIT_BUTTON = '[data-qa-id=meeting-name-submit]';
+
+/**
+ * A page object for interacting with the input field for entering a meeting
+ * name to join.
+ */
+class MeetingInput extends PageObject {
+    /**
+     * Initializes a new {@code MeetingInput} instance.
+     *
+     * @param {Object} driver - The webdriver.io browser instance which can
+     * interact with Spot.
+     */
+    constructor(driver) {
+        super(driver);
+
+        this.rootSelector = INPUT;
+    }
+
+    /**
+     * Sets the input value.
+     *
+     * @returns {void}
+     */
+    setMeetingName(meetingName) {
+        this.waitForVisible();
+        this.driver.setValue(INPUT, meetingName);
+    }
+
+    /**
+     * Clicks on the submit button.
+     *
+     * @returns {void}
+     */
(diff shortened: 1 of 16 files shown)


virtuacoplenny:
Force pushed to rebase on top of latest master, which now has PR #15 merged in.

ui-tests/page-objects/meeting-input.js
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * A page object for interacting with the input field for entering a meeting
+ * name to join.
+ */
+class MeetingInput extends PageObject {
+    /**
+     * Initializes a new {@code MeetingInput} instance.
+     *
+     * @param {Object} driver - The webdriver.io browser instance which can
```
zbettenbuk: Don't you want to use inheritdoc here (and other places) to avoid duplications?

ui-tests/page-objects/meeting-input.js
```diff
@@ -0,0 +1,54 @@
...
+    constructor(driver) {
+        super(driver);
+
+        this.rootSelector = INPUT;
+    }
+
+    /**
+     * Sets the input value.
+     *
+     * @returns {void}
```
zbettenbuk: Isn't the param part missing? Or is it another ruleset that doesn't require having that?

ui-tests/page-objects/meeting-page.js
```diff
@@ -0,0 +1,39 @@
+const { URL } = require('url');
```
zbettenbuk: May be a stupid question: don't we use flow for these files? Or is it too early for that?

virtuacoplenny: I'd say not using flow has more to do with my own inabilities coupled with the need to quickly blast through prototyping. I'm assuming a lot of the existing code will change as a more refined version emerges, and afterwards I'd want to add flow for consistency with jitsi-meet. I didn't even really want jsdocs right now but thought they'd be helpful in case someone else that's not a js person was going to make changes.

ui-tests/page-objects/meeting-input.js
virtuacoplenny: Looks like it has something to do with the eslint config having extended both plugin:react/recommended eslint-config-jitsi. I added the param. I'll try to find out why the problems occurs and fix the config and the subsequent errors in another PR.

virtuacoplenny:
Rebasing onto latest master in prep for merge.

PR #26, opened 2018-11-29
ref(keyboard-nav): redo with react components

virtuacoplenny:
No more custom arrow navigation. Instead rely on tabIndex and
tabbing for now.

This will probably get re-written again.


virtuacoplenny:
Spec no longer includes arrow key navigation so this refactoring is not needed.

PR #27, opened 2018-12-01
fix(calendar): change when initial fetch is marked as done, rename vars

virtuacoplenny:


virtuacoplenny:
We has discussed moving the logic of fetching calendars to the calendar list itself as part of this next set of changes. But I'm actually going to resist doing that for now and revisit next week. I like that the calendar list right now has no state other than what is passed to it, and almost all the brains of the app is in the src/views files, versus the individual components.

virtuacoplenny:
Merged into the beta branch instead of master.

PR #29, opened 2018-12-01
ref(loaders): pass service into children

virtuacoplenny:

PR #31, opened 2018-12-02
ref(remote-control): move popup handling management to class 

virtuacoplenny:
Built on top of #29.

Instead of remote control window opening logic being in the calendar view, move it to its own class so it can be aware of the popup and interact with it. This is done to separate calendar view from popup handling, in case the popup becomes more than a debug feature.


virtuacoplenny:
Spec had changed again a couple weeks ago and now popup remote control isn't required.

PR #37, opened 2018-12-26
feat(calendar): outlook integration

virtuacoplenny:


virtuacoplenny:
There's a decent amount of duplication now between jitsi-meet and jitsi-meet-spot. This wasn't the case a year ago when spot had calendar integration and meet didn't, but now both will have integration. This brings up the question again of whether or not spot should live outside of jitsi-meet. At the least, likely the shared logic for handling the Microsoft auth token should be encapsulated in a common place.

saghul:
We do have a js-utils package, maybe some stuff could live there? Or maybe some new Jitsi-Meet-common package?

At any rate, how much is the duplication? These days I lean more towards “a little duplication” than “a little dependency”.

virtuacoplenny:
The logic for handling ms auth tokens and renewal is copied pasted with the main difference being instead of storing stuff in redux I store it in local storage so its knowledge is independent from the spot app; so most of src/calendars/microsoft-client-api/api.js is copied. I was also thinking the logic for handling tokens should be part of the microsoft-graph-client package, which I don't control, or it should be a separate package altogether. I'm hesitant to dump it into js-utils because I don't understand what js-utils even becomes then--maybe its name really should have been jitsi-meet-common.

virtuacoplenny:
I'm moving all my changes to a beta branch to avoid the PR process for each change. Also, I'm just gonna keep copying/pasting code until I reach a point where I can clean the duplication; it'll be a good problem to have if that's really my top concern.

PR #40, opened 2018-12-30
fix(setup): check resource type when filtering rooms

virtuacoplenny:
It may be that conference rooms configured in Google
are not set to the proper resource category but have
a manual resource type set.


virtuacoplenny:
Abandoning creating PRs for spot.

PR #52, opened 2019-01-18
feat(screenshare): support remote wireless screensharing

virtuacoplenny:
Relies on https://github.com/jitsi/jitsi-meet/pull/3809 and https://github.com/jitsi/lib-jitsi-meet/pull/857.

### DIFF
--- a/src/features/meeting-frame/meeting-frame.js
+++ b/src/features/meeting-frame/meeting-frame.js
@@ -85,6 +85,8 @@ export default class MeetingFrame extends React.Component {
         this._jitsiApi.addListener(
             'feedbackSubmitted', this.props.onMeetingLeave);
         this._jitsiApi.addListener(
+            'proxyConnectionEvent', this._onSendMessageToRemoteControl);
+        this._jitsiApi.addListener(
             'readyToClose', this.props.onMeetingLeave);
         this._jitsiApi.addListener(
             'screenSharingStatusChanged', this._onScreenshareChange);
@@ -229,6 +231,22 @@ export default class MeetingFrame extends React.Component {
         this._isScreensharing = on;
 
         remoteControlService.notifyScreenshareStatus(on);
+    }
+
+    /**
+     * Passes a message from {@code JitsiMeetExternalAPI} to a specified
+     * remote control.
+     *
+     * @param {Object} event - The object holding information on what and how to
+     * send the message.
+     * @param {string} event.to - The jid of the remote control which should
+     * receive the message.
+     * @param {Object} event.data - Details of the message.
+     * @private
+     * @returns {void}
+     */
+    _onSendMessageToRemoteControl({ to, data }) {
+        remoteControlService.sendMessageToRemoteControl(to, data);
     }
 
     /**
--- a/src/features/remote-control-menu/buttons/wireless-screenshare-button.js
+++ b/src/features/remote-control-menu/buttons/wireless-screenshare-button.js
@@ -14,10 +14,14 @@ import styles from '../remote-control-menu.css';
  * @returns {ReactElement}
  */
 export default function WirelessScreenshareButton(props) {
-    const { isProxyConnectionActive, isScreensharing, onClick } = props;
+    const {
+        isScreensharing,
+        isWirelessScreenshareConnectionActive,
+        onClick
+    } = props;
     let buttonText;
 
-    if (isProxyConnectionActive && !isScreensharing) {
+    if (isWirelessScreenshareConnectionActive && !isScreensharing) {
         buttonText = 'Cancel screensharing';
     } else if (isScreensharing) {
         buttonText = 'Stop screensharing';
@@ -38,7 +42,7 @@ export default function WirelessScreenshareButton(props) {
 }
 
 WirelessScreenshareButton.propTypes = {
-    isProxyConnectionActive: PropTypes.bool,
     isScreensharing: PropTypes.bool,
+    isWirelessScreenshareConnectionActive: PropTypes.bool,
     onClick: PropTypes.func.isRequired
 };
--- a/src/features/remote-control-menu/buttons/wireless-screenshare-button.js
+++ b/src/features/remote-control-menu/buttons/wireless-screenshare-button.js
@@ -2,7 +2,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import { Button } from 'features/button';
-
+import { LoadingIcon } from 'features/loading-icon';
 import styles from '../remote-control-menu.css';
 
 /**
@@ -19,14 +19,14 @@ export default function WirelessScreenshareButton(props) {
         isWirelessScreenshareConnectionActive,
         onClick
     } = props;
-    let buttonText;
+    let buttonContent;
 
     if (isWirelessScreenshareConnectionActive && !isScreensharing) {
-        buttonText = 'Cancel screensharing';
+        buttonContent = <LoadingIcon color = 'black' />;
     } else if (isScreensharing) {
-        buttonText = 'Stop screensharing';
+        buttonContent = 'Stop screensharing';
     } else {
-        buttonText = 'Start wireless screensharing';
+        buttonContent = 'Start wireless screensharing';
     }
 
     return (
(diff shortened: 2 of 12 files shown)


src/features/meeting-frame/meeting-frame.js
```diff
@@ -78,6 +78,8 @@ export default class MeetingFrame extends React.Component {
             'audioMuteStatusChanged', this._onAudioMuteChange);
         this._jitsiApi.addListener(
             'feedbackSubmitted', this.props.onMeetingLeave);
+        this._jitsiApi.addListener(
+            'proxyConnectionEvent', this._onSendMessageToRemoteControl);
```
bbaldino: reading this, i thought it was the handler for an incoming (iframe host -> iframe) event, but it looks like this is an outgoing (iframe -> iframe host) event?

src/features/remote-control-menu/buttons/wireless-screenshare-button.js
```diff
@@ -14,10 +14,14 @@ import styles from '../remote-control-menu.css';
...
-    const { isProxyConnectionActive, isScreensharing, onClick } = props;
+    const {
+        isScreensharing,
+        isWirelessScreenshareConnectionActive,
+        onClick
+    } = props;
     let buttonText;
 
-    if (isProxyConnectionActive && !isScreensharing) {
+    if (isWirelessScreenshareConnectionActive && !isScreensharing) {
```
bbaldino: what does `isScreensharing` represent here?  If `isScreensharing` is false we show 'cancel screensharing` ?

Oh, does this test represent an state in between starting the connection and actually transmitting?

src/features/remote-control-menu/remote-control-menu.js
```diff
@@ -106,4 +121,24 @@ export default class RemoteControlMenu extends React.Component {
...
         remoteControlService.setVideoMute(!this.props.videoMuted);
     }
+
+    /**
+     * Changes the current wireless screensharing state.
+     *
+     * @private
+     * @returns {void}
+     */
+    _onSetWirelessScreensharing() {
```
bbaldino: this says `set` but really it's a toggle, no?

src/remote-control/process-update-delegate.js
```diff
@@ -132,6 +148,50 @@ export default class ProcessUpdateDelegate {
...
+            // Jitsi-Meet.
+            this._sendEventIfInMeeting({
+                data,
+                from
+            });
+
+            break;
+        }
+
+        return Promise.resolve(ack);
```
bbaldino: we return a promise here just to make it easier for the caller to chain some things together?  nothing actually async here (that the ack is based on, at least) right?

src/remote-control/process-update-delegate.js
```diff
@@ -146,7 +206,32 @@ export default class ProcessUpdateDelegate {
...
+                const iq = $iq({ type: 'set' })
+                    .c('jingle', {
+                        xmlns: 'urn:xmpp:jingle:1',
+                        action: 'unavailable'
+                    })
+                    .c('details')
+                    .t('unavailable')
+                    .up();
+                const stringifiedIq
+                    = new XMLSerializer().serializeToString(iq.nodeTree);
```
bbaldino: is this what we use in the lib?  i don't remember seeing it...i assumed the strophe types had a member method to serialize

src/remote-control/process-update-delegate.js
```diff
@@ -177,11 +262,75 @@ export default class ProcessUpdateDelegate {
...
+     * remote control and a Jitsi-Meet meeting.
+     *
+     * @param {string} spotId - The jid of the Spot to use as a signaling layer
+     * for establishing a screenshare connection. Messages go to Spot for it
+     * to pass into the meeting.
+     * @param {string} roomFullJId - The jid of the remote control that wants
+     * to establish a {@code ScreenshareConnection}.
+     * @param {Object} screenshareConnection - The instance of
+     * {@code ScreenshareConnection} to be used.
+     * @returns {Promise}
```
bbaldino: more detail on when the promise resolves (and with what)?

src/remote-control/process-update-delegate.js
```diff
@@ -177,11 +262,75 @@ export default class ProcessUpdateDelegate {
...
+        this._screenshareConnection = screenshareConnection;
+
+        this._store.dispatch(updateSpotState({
+            isWirelessScreenshareConnectionActive: true
+        }));
+
+        return screenshareConnection.startScreenshare(
+            spotId,
+            roomFullJId
+        ).catch(error => {
```
bbaldino: dumb question: if you catch here, does the failure still get propagated up to the promise that was returned? (i.e., if the caller has a catch?) or do we not want that?

src/remote-control/process-update-delegate.js
```diff
@@ -194,23 +343,58 @@ export default class ProcessUpdateDelegate {
...
+
+    /**
+     * Invokes the {@code sendProxyConnectionEvent} method on the
+     * {@code JitsiMeetExternalApi} if it exists.
+     *
+     * @param {Object} event - The event to pass into the api.
+     * @private
+     * @returns {void}
+     */
+    _sendEventIfInMeeting(event) {
```
bbaldino: this is a total bullshit comment, but for some reason this method is throwing me off.  i would've expected it to be inverted:
```
const meetingApi = this._getMeetingApiIfInMeeting();

if (!meetingApi) {
  logger.error('Failed to send proxy event.');
  return;
}

meetingApi.sendProxyConnectionEvent(event);
```

src/remote-control/remote-control-service.js
```diff
@@ -201,6 +218,56 @@ class RemoteControlService {
...
+            const connection = new ScreenshareService({
+
+                /**
+                 * Callback invoked when the connection has been closed
+                 * automatically. Triggers cleanup of
+                 * {@code ScreenshareService}.
+                 *
+                 * @returns {void}
+                 */
+                onConnectionClosed: () => this._delegate.stopScreenshare(),
```
bbaldino: i don't have a full understanding of the role of the delegate/what's in there vs what's in this screenshare service.  maybe you can explain it to me more.

src/remote-control/screenshare-connection.js
```diff
@@ -0,0 +1,127 @@
...
+        this.options = options;
+
+        /**
+         * Whether or not the {@code ProxyConnectionService} has an active
+         * connection. Used to prevent connection events from firing after the
+         * connection has been stopped.
+         *
+         * @type {boolean}
+         */
+        this._isStopped = true;
```
bbaldino: `_isActive`?  the weird thing about `_isStopped` is you kinda get a double negative in the `if (!_isStopped)` case

src/features/remote-control-menu/remote-control-menu.js
virtuacoplenny: You are right. I think I followed the naming convention I had in this file previously, where everything was set for some reason. I can change this instance and change the others in a future pr.

src/remote-control/process-update-delegate.js
virtuacoplenny: Correct, this is for caller chaining, but you are right that there isn't a need to do anything async. What I bet happened is at some point I wanted to do promise.reject but thinking now that wouldn't make sense because the rejection/error would be set during ack creation time. I should probably audit the promise use here.

src/remote-control/process-update-delegate.js
virtuacoplenny: strophe does have toString on its builder I should use.

src/remote-control/process-update-delegate.js
virtuacoplenny: It won't get propagated up. At some point if I don't catch  itand handle it an error will be logged about an uncaught promise rejection. I can double check swallowing the error is in fact what I want.

src/remote-control/process-update-delegate.js
virtuacoplenny: you probably expected that because that's my normal style, and maybe a normal jitsi style(?) to handle errors first as the exception. I should change this.

src/remote-control/screenshare-connection.js
virtuacoplenny: Fair. You are correct; state should be positive to avoid the double negative.

virtuacoplenny:
Force pushed to resolve conflicts with the beta branch.

virtuacoplenny:
Force pushing to resolve conflicts. Changed external_api script src to beta now that the jitsi-meet side is deployed.

PR #112, opened 2019-02-21
Adds retry logic to logs publisher

paweldomas:


spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -1,18 +1,60 @@
...
+     * How many times it'll try to send a log batch.
+     * @type {number}
+     */
+    static MAX_ATTEMPTS = 4;
+
+    /**
+     * Gets next timeout using the full jitter pattern.
+     *
+     * @param {number} attempt - The attempt number. On the first retry
+     * the value is 2.
```
paweldomas: this doc is wrong, on the first retry the value is 1

paweldomas: this is probably confusing I may need to fix that attempt/retry wording

spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -54,19 +96,76 @@ export default class PostToEndpoint {
...
+                if (!ok && status >= 500 && status < 600) {
+                    // Retry on server errors and network failures which
+                    // is a regular fetch reject case.
+                    throw new Error(`Request error - status: ${status}`);
+                } else if (ok) {
+                    // Try sending the next request from the queue
+                    this._sendLogs();
+                } else {
+                    // eslint-disable-next-line no-console
+                    console.warn(`Dropping log request, status: ${status}`);
```
virtuacoplenny: Should you still call send logs here in case there’s something queued?

spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -54,19 +96,76 @@ export default class PostToEndpoint {
...
+     * Tries to send first log request from the requests queue. In case of
+     * success it will keep taking requests of the queue until it's empty.
+     * On failure it will retry the request with a delay until
+     * the {@link PostToEndpoint#MAX_ATTEMPTS} is reached.
+     *
+     * @returns {void}
+     * @private
+     */
+    _sendLogs() {
+        const request = this._requestsQueue.shift();
```
virtuacoplenny: Thought of a race here: if you pull out the logs now the array could be empty while the fetch is in flight. If logs get queued then another fetch could get fired.

spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -1,18 +1,60 @@
+
+/**
+ * The structure holds the meta data about the log requests waiting in the queue
+ * to be sent.
+ */
+class PostLogsRequest {
```
virtuacoplenny: This class was a really good idea. Really good way to mentally keep together some variables.

spot-client/src/common/logger/post-to-endpoint.js
paweldomas: Yup, good catch. So maybe the request should not be removed from the queue until it gets processed ?

spot-client/src/common/logger/post-to-endpoint.js
paweldomas: yes

spot-client/src/common/logger/post-to-endpoint.js
virtuacoplenny: Yeah. And that way you also don't have to re-add the payload if it fails.

spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -138,29 +137,28 @@ export default class PostToEndpoint {
...
             .catch(error => {
                 // eslint-disable-next-line no-console
-                console.error(`Log request error, attempt: ${attempt}`, error);
+                console.error(`Log request error, attempt: ${retry + 1}`, error);
 
-                if (attempt < PostToEndpoint.MAX_ATTEMPTS) {
-                    // Put the current request back at the front of the queue
-                    this._requestsQueue.unshift(
-                        new PostLogsRequest(attempt + 1, events));
+                if (retry < PostToEndpoint.MAX_RETRIES) {
```
paweldomas: forgot about one thing: now the request must be removed also when the retry limit is exceeded

spot-client/src/common/logger/post-to-endpoint.js
```diff
@@ -54,19 +96,83 @@ export default class PostToEndpoint {
...
+     * On failure it will retry the request with a delay until
+     * the {@link PostToEndpoint#MAX_RETRIES} is reached.
+     *
+     * @returns {void}
+     * @private
+     */
+    _sendLogs() {
+        const request = this._requestsQueue.length && this._requestsQueue[0];
+
+        if (request) {
```
virtuacoplenny: Style comment: it might be easier to read if this were an early return because that's an exceptional/error case, plus it prevents having to remember the following code is within an if.

PR #116, opened 2019-02-22
feat(meeting-frame): display an overlay to hide Jitsi's feedback prompt

paweldomas:


spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -142,9 +145,12 @@ export default class MeetingFrame extends React.Component {
      */
     render() {
         return (
-            <div
-                className = 'meeting-frame'
-                ref = { this._setMeetingContainerRef } />
+            <div className = 'meeting-frame'>
```
paweldomas: I don't like this frame duplication, but I don't know any better. Is it possible to return an array here ?

paweldomas: I see it can be replaced with <> to use React fragments, but I'll wait with changing that to hear from you if it's ok to do it in meeting-frame at all

virtuacoplenny: Yeah, it's okay. We run fast and loose here, but more importantly using a fragment is idiomatic anyway.

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -240,6 +249,9 @@ export default class MeetingFrame extends React.Component {
      */
     _onMeetingLoaded() {
         this._meetingLoaded = true;
+        this.setState({
+            feedbackDisplayed: false
+        });
```
virtuacoplenny: Is false needed here?

paweldomas: I don't was just doing it for safety if we'd be reloading it or something. I'm not sure if it makes sense ?

PR #119, opened 2019-02-22
feat(xmpp-connection): set 'disableFocus' to true

paweldomas:
Do not talk to Jicofo when joining the room.

### DIFF
--- a/spot-client/src/common/remote-control/xmpp-connection.js
+++ b/spot-client/src/common/remote-control/xmpp-connection.js
@@ -182,7 +182,7 @@ export default class XmppConnection {
             return this.room;
         }
 
-        this.room = this.xmppConnection.xmpp.createRoom(roomName, {});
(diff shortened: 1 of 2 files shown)


spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -193,16 +193,16 @@ export default class XmppConnection {
...
-     * @returns {void}
+     * @returns {Promise} - A Promise resolved when libjitsi-meet's ChatRoom.join method is resolved
+     * which is not exactly equal with being in the muc already.
      */
     _joinMuc(lock) {
-        if (lock) {
-            this.room.join(lock);
-        }
-
-        // send presence manually to avoid focus joining
```
virtuacoplenny: Oh that's funny. I guess I thought about this at one point somehow.

paweldomas:
Requires https://github.com/jitsi/lib-jitsi-meet/pull/876 otherwise spot will not join the MUC, because the sendPresence call is removed.

PR #124, opened 2019-02-24
Split up redux files

virtuacoplenny:


virtuacoplenny:
Maybe this makes things more organized...?

PR #132, opened 2019-02-28
Nest config in dist/config

virtuacoplenny:
This will make it easier to override for jitsi-meet's spot-client deployments.

### DIFF
--- a/spot-client/config.js
+++ b/spot-client/config.js
@@ -124,10 +124,11 @@ window.JitsiMeetSpotConfig = {
      * the jitsi UI.
      */
     XMPP_CONFIG: {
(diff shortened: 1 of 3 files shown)


spot-client/config.js
```diff
@@ -124,10 +124,11 @@ window.JitsiMeetSpotConfig = {
      * the jitsi UI.
      */
     XMPP_CONFIG: {
-        bosh: process.env.XMPP_BOSH || '',
+        bosh: process.env.XMPP_BOSH || 'https://beta.meet.jit.si/http-bind',
```
aaronkvanmeerten: Was it intentional to set these to beta?  Might it not make sense for the open source to use meet.jit.si by default instead of beta?

PR #137, opened 2019-03-01
[hold merge] do redirection of all paths not matching existing files, back to /

aaronkvanmeerten:
This change causes nginx to redirect all requests back to /, similar to how jitsi-meet operates
If a file exists, it is served, otherwise we serve index.html for any request made.


aaronkvanmeerten:
I vote we don't merge it until after spot is URL aware.  This is just an aid to whomever implements that.

virtuacoplenny:
We'll have to point this PR to master as beta is now the old master and master is the new master.

paweldomas:
Moved it here: https://github.com/jitsi/jitsi-meet-spot/pull/155

PR #142, opened 2019-03-04
ref: rewrite data flows between xmpp and app-state

virtuacoplenny:
- Add manual test plan doc for future testing reference.
- Renamed redux state remoteControl to spotTv to better signify
  it represents the current presence/state of Spot-TV.
- Removed store of meetingApi in redux. Instead the meeting
  Frame sets callbacks on the remote control service to get
  presence and iq updates. 
- Create a RemoteControlServiceSubscriber. The changes make
  Spot-TV update its redux, and then presence is updated;
  previously presence was updated and the presence update
  would trigger a redux update. This makes it so local
  known state can get stale if an error occurs in the
  XMPP connection. The subscriber will automatically
  trigger a presence update as Spot-TV updates redux.
- Move screensharing handling into RemoteControlService, as
  suggested while wireless screensharing from home was
  being implemented.
- Consolidate all RemoteControlService notify___Status
  methods into an updateStatus method.
- Spot-Remotes update Spot-TV through a callback in
  withRemoteControl instead of in processUpdateDelegate.
- Spot-TV handles commands and messages through callbacks
  registered by the meeting frame view instead of
  through processUpdateDelegate.


virtuacoplenny:
Functionality I know I'm missing:
- [ ] cleaning up screenshare if the meeting fails to join
- [ ] doing something with the isWirelessScreenshareConnectionActive state
- [ ] a new data flow chart

spot-client/src/common/app-state/spot-tv/selectors.js
```diff
@@ -0,0 +1,71 @@
+/**
+ * A selector which returns the last known room lock set on the Spot-TV's MUC,
+ * which must be entered to join the MUC.
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {string}
+ */
+export function getCurrentLock(state) {
```
paweldomas: When I first time see "lock" Im thinking about this like about a boolean, whether the room is locked or not. Would it be better to use password ? The XMPP spec names locked rooms "password protected", so it feels closer to what it really is

paweldomas: not sure if just "password" or "mucPassword"

paweldomas: or roomPassword

spot-client/src/common/remote-control/remote-control-service-subscriber.js
```diff
@@ -0,0 +1,40 @@
...
+
+        if (newSpotTvState === this._previousSpotTvState) {
+            return;
+        }
+
+        remoteControlService.updateStatus(newSpotTvState);
+
+        if (this._previousSpotTvState.inMeeting !== newSpotTvState.inMeeting
+            && newSpotTvState.inMeeting) {
+            remoteControlService.startAnyDeferredWirelessScreenshare();
```
paweldomas: I like that

spot-client/src/common/app-state/spot-tv/selectors.js
virtuacoplenny: Okay, we'll say mucPassword. I think at some point it won't be in redux, once the backend stuff gets made, so specifying it's for a muc is good. Plus it differentiates between a conference room, meeting room, and muc room.

virtuacoplenny: I'm gonna do this in a follow up PR once this PR lands and is stable. The naming is actually from existing logic and the term "lock" is used in a lot of places.

virtuacoplenny:
Force pushed to resolve conflicts in:
spot-client/src/common/remote-control/remote-control-service.js
spot-client/src/index.js

Conflicts were related to adding a global debugging helper to registering global variables.

PR #145, opened 2019-03-05
fix(app): prevent zoom and scroll

virtuacoplenny:


in-room-controller/src/remote-control.js
```diff
@@ -16,6 +17,19 @@ export default class RemoteControl extends React.PureComponent {
         url: PropTypes.string
     };
 
+    _preventWebViewZoomScript = `
+        var metaTag = document.getElementsByName('viewport')[0];
+        if (metaTag) {
+            var content = metaTag.getAttribute('content') || '';
+            var preventZoom = 'maximum-scale=1.0, user-scalable=0';
```
paweldomas: I'm reading on the internets that the values for `user-scalable` are either `yes` or `no`. Is that not the case from your experience ?

virtuacoplenny: I was seeing both, but the examples I saw were from old blog posts. Instead of adding something that could be incorrect I'll stick with just maximum-scale and then re-visit as needed. But yes it was working.

paweldomas: so is it okay with the user-scalable ?

virtuacoplenny: Yeah it's fine without the user-scalable when I tested. It's fine with either-or.

PR #147, opened 2019-03-05
Fix usb dongle reconnect

paweldomas:


spot-client/src/spot-tv/wired-screenshare-service/video-change-listener.js
```diff
@@ -307,4 +313,13 @@ export default class VideoChangeListener {
...
     }
+
+    /**
+     * Implements toString for logging convenience.
+     *
+     * @returns {string}
+     */
+    toString() {
+        return `VideoChangeListener[deviceLabel=${this._deviceLabel}]`;
+    }
```
virtuacoplenny: Nice. I've never overridden toString and what a great idea.

PR #153, opened 2019-03-06
fix(controller): add placeholder icons

virtuacoplenny:


paweldomas:
okay, I guess now it's the time I finally build the iOS app and give it a try

PR #154, opened 2019-03-06
say byte to hash routing

paweldomas:


spot-client/src/common/utils/window-handler.js
```diff
@@ -10,7 +10,7 @@ export default {
      * @returns {string}
      */
     getBaseUrl() {
```
paweldomas: I feel like this change violates the definition, but not sure

virtuacoplenny: seems okay. window handler only exists to hide window.location related accessing for the unit tests that don't exist.

spot-client/src/spot-tv/ui/views/home.js
```diff
@@ -189,7 +189,7 @@ export class Home extends React.Component {
      */
     _onOpenRemote() {
         const baseUrl = windowHandler.getBaseUrl();
-        const url = `${baseUrl}#/?code=${this.props.joinCode.toUpperCase()}`;
+        const url = `${baseUrl}/${this.props.joinCode.toUpperCase()}`;
```
virtuacoplenny: We should hide this one day in some kind of routing helper. Did you happen to find a decent place to shove it?

spot-client/src/spot-remote/ui/views/join-code-entry.js
```diff
@@ -65,15 +65,16 @@ export class JoinCodeEntry extends React.Component {
         this.props.dispatch(setJoinCode(''));
         remoteControlService.disconnect();
 
-        const queryParams = new URLSearchParams(this.props.location.search);
-        const code = queryParams.get('code');
+        const { pathname } = this.props.location;
+        const codeMatch = pathname.match(new RegExp('^/(\\w{6})$'));
+        const code = codeMatch && codeMatch[1];
```
virtuacoplenny: So the default already goes down to the remote control page?

paweldomas: yes, here: https://github.com/jitsi/jitsi-meet-spot/blob/master/spot-client/src/app.js#L158

It doesn't have a path defined, that's why

spot-client/src/spot-tv/ui/views/home.js
paweldomas: No, but routing helper sounds good to me. Do you want me to create one now ?

virtuacoplenny: Nah, not for this PR.

PR #156, opened 2019-03-07
feat(controller): add fastlane support

virtuacoplenny:


in-room-controller/ios/fastlane/report.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
saghul: add this to gitignore

PR #158, opened 2019-03-07
fix(screenshare): notify when screenshare input no longer set

virtuacoplenny:
The edge case is a Spot-TV that has been set up with a screenshare
dongle and goes through setup again without selecting a dongle. In
that case the Spot-Remotes will continue to believe the Spot-TV has
a dongle connected. So instead update state when the dongle is
no longer set.


virtuacoplenny:
Force pushed to rebase onto latest master and squash commits into one.

virtuacoplenny:
Force pushed to resolve conflicts around moving enumerate devices logic into avUtils.

PR #161, opened 2019-03-09
feat(meeting): hide filmstrip when screensharing in lonely call

virtuacoplenny:
Relies on https://github.com/jitsi/jitsi-meet/pull/3974. One challenge for the future will be if the Spot-Remote ever has a filmstrip toggle button; there will be questions on how autohiding can live with manual hiding.

### DIFF
--- a/spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
+++ b/spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
@@ -46,17 +46,29 @@ export class MeetingFrame extends React.Component {
     constructor(props) {
         super(props);
 
+        /**
+         * The external api exposes toggle actions. Store various state to
+         * prevent execution of remote commands when the meeting is already in
+         * the desired state.
+         */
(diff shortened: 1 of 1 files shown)


spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -324,6 +357,52 @@ export class MeetingFrame extends React.Component {
...
+     * Adds a participant from the list of known current participants in the
+     * meeting.
+     *
+     * @param {Object} event - An object with participant information.
+     * @param {string} event.id - The jitsi-meet ID of the participant.
+     * @param {string} even.displayName - The current name of the participant.
+     * @private
+     * @returns {void}
+     */
+    _onParticipantJoined({ id, displayName }) {
```
paweldomas: I would probably prefer to have this kind of logic in a class separate from UI component. This way it makes it less likely for someone else to duplicate this same logic again, but because it's just the two of us right now it's not a big deal.

PR #167, opened 2019-03-13
feat: enable TURN

paweldomas:
Will pass JitsiConnection to the ProxyConnectionService, so that it can use it to fetch the TURN credentials from Prosody. Those will be used in the P2P screensharing connection.

### DIFF
--- a/spot-client/src/common/remote-control/remote-control-service.js
+++ b/spot-client/src/common/remote-control/remote-control-service.js
@@ -449,6 +449,11 @@ class RemoteControlService {
                 this._wirelessScreensharingConfiguration || {},
 
             /**
+             * The {@code JitsiConnection} instance will be used to fetch TURN credentials.
+             */
+            jitsiConnection: this.xmppConnection.getJitsiConnection(),
+
+            /**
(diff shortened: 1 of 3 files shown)


spot-client/config.js
```diff
@@ -139,6 +139,9 @@ window.JitsiMeetSpotConfig = {
             domain: process.env.XMPP_HOSTS_DOMAIN || 'meet.jit.si',
             muc: process.env.XMPP_HOSTS_MUC_URL
                 || 'conference.meet.jit.si'
+        },
+        p2p: {
+            useStunTurn: true
```
virtuacoplenny: We'll have to update the deployment config too (when we're about to deploy) because right now with the current setup the deployment config overrides the config completely.

paweldomas: oh okay

paweldomas: will this work better if places in the defaults then ?

virtuacoplenny: My other PR about config refactoring could help but I want to make sure we're in agreement on that one. But yeah, I think right now adding default true would outside of config would be really good.

PR #169, opened 2019-03-14
Misc webdriver config tweaks

virtuacoplenny:


spot-webdriver/wdio.conf.js
```diff
@@ -2,6 +2,10 @@
 
 const path = require('path');
 
+const LOG_LEVEL = process.env.LOG_LEVEL || 'warn';
+const PATH_TO_FAKE_VIDEO
+    = path.resolve(__dirname, 'resources/static-image.y4m');
```
paweldomas: Maybe it would be good to add the `static-image` file in this commit as well ?

virtuacoplenny: I agree.

PR #173, opened 2019-03-15
feat(tests): basic wireless screensharing flows

virtuacoplenny:
- Check wireless screensharing can be started in meeting and
  out of meeting, both with wired enabled and not enabled.
- Add a global afterEach to clear persisted app state between
  tests.

### DIFF
--- a/spot-webdriver/page-objects/admin-page.js
+++ b/spot-webdriver/page-objects/admin-page.js
@@ -1,0 +1,72 @@
+const constants = require('../constants');
+const PageObject = require('./page-object');
+
+const ADMIN_VIEW = '[data-qa-id=admin-view]';
+const CHANGE_SCREENSHARE_BUTTON = '[data-qa-id=admin-change-screenshare]';
+const EXIT_BUTTON = '[data-qa-id=admin-exit]';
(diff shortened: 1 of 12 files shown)


spot-webdriver/page-objects/admin-page.js
```diff
@@ -0,0 +1,72 @@
...
+        button.click();
+    }
+
+    /**
+     * Proceeds directly to the admin view of a Spot-TV.
+     *
+     * @returns {void}
+     */
+    visit() {
+        this.driver.url(constants.SPOT_TV_ADMIN_URL);
```
paweldomas: It seems that the `visit` method is duplicated across the `PageObject`s implementations

spot-webdriver/specs/wireless-screenshare.spec.js
```diff
@@ -1,32 +1,106 @@
+const constants = require('../constants');
 const remoteControlConnect = require('../flow-utils/remote-control-connect');
```
paweldomas: 'remoteControlConnect' has been removed, so it needs to be updated now

virtuacoplenny:
Force pushed to fix conflicts.

paweldomas:
There are some lint errors

PR #176, opened 2019-03-17
feat(setup): add way to set custom name and avatar

virtuacoplenny:
No designs for this yet but add in the hooks for it.

### DIFF
--- a/spot-client/src/common/app-state/setup/selectors.js
+++ b/spot-client/src/common/app-state/setup/selectors.js
@@ -1,4 +1,38 @@
 /**
+ * A selector which returns the configured url to use for the Spot-TV avatar
+ * which should be displayed in meetings.
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {boolean}
+ */
+export function getAvatarUrl(state) {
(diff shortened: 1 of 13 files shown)


spot-client/src/common/app-state/setup/selectors.js
```diff
@@ -30,3 +31,24 @@ export function getMeetingOptions(state) {
...
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {boolean}
+ */
+export function isSetupComplete(state) {
+    return state.setup.completed;
+}
+
+/**
+* A selector which returns whether or not the current client is a Spot
```
paweldomas: This is a bit confusing when Im reading it knowing about Spot-TV and Spot-Remote. Is it about Spot-TV ? Is it the flag that's read from presence ?

paweldomas: This method was there already is just moved around, right ?

virtuacoplenny: Yeah, I moved it; didn't modify its contents.

PR #179, opened 2019-03-18
Audio muted test

paweldomas:


spot-webdriver/user/spot-session.js
```diff
@@ -0,0 +1,65 @@
...
+        meetingInput.submitMeetingName(testMeetingName);
+
+        const meetingPage = this.spotTV.getMeetingPage();
+
+        meetingPage.waitForVisible();
+
+        expect(meetingPage.getMeetingName()).toBe(testMeetingName);
+    }
+}
+
```
virtuacoplenny: Do you think remotes and TVs should be obtained through the session?

spot-webdriver/user/spot-session.js
```diff
@@ -0,0 +1,65 @@
...
+        const testMeetingName = meetingName ? meetingName : `ui-test-${Date.now()}`;
+        const meetingInput = remoteControlPage.getMeetingInput();
+
+        meetingInput.submitMeetingName(testMeetingName);
+
+        const meetingPage = this.spotTV.getMeetingPage();
+
+        meetingPage.waitForVisible();
+
+        expect(meetingPage.getMeetingName()).toBe(testMeetingName);
```
virtuacoplenny: I'm not sure how I feel about a jasmine expectation being defined in session. I'm not against it but I'm not used expects being located outside the spec file.

paweldomas: I will look into that

spot-webdriver/specs/in-meeting.spec.js
```diff
@@ -0,0 +1,25 @@
...
+    meetingRemoteControlPage.waitForVisible();
+
+    it('can toggle audio mute', () => {
+        meetingRemoteControlPage.muteAudio();
+
+        meetingRemoteControlPage.waitForAudioMutedStateToBe(true);
+
+        meetingRemoteControlPage.unmuteAudio();
+
+        meetingRemoteControlPage.waitForAudioMutedStateToBe(false);
```
virtuacoplenny: I wonder if waiting for mute state should be part of the unmute/mute action. The downside I see is tests aren't as explicit about what is being tested.

virtuacoplenny: This applies to anything I've already written too.

PR #186, opened 2019-03-20
feat(feedback-form): add inactivity timeout

paweldomas:


spot-client/src/spot-remote/ui/components/remote-control-menu/feedback-form.js
```diff
@@ -59,6 +85,19 @@ export default class FeedbackForm extends React.Component {
...
     }
 
+    /**
+     * Cancels the inactivity timeout.
+     *
+     * @private
+     * @returns {void}
+     */
+    _cancelInactivityTimeout() {
+        if (this._timeout) {
```
virtuacoplenny: You don't need an truthy check for clearInterval/clearTimeout. The functions take in maybe anything and won't error.

spot-client/src/spot-remote/ui/components/remote-control-menu/feedback-form.js
```diff
@@ -175,4 +211,38 @@ export default class FeedbackForm extends React.Component {
...
+    /**
+     * Submits the feedback currently stored in the state (whatever it is).
+     *
+     * @param {boolean} timeout - Whether or not the feedback is being submitted as a result of
+     * the inactivity timeout.
+     * @private
+     * @returns {void}
+     */
+    _submitFeedback(timeout) {
+        logger.log(`Feedback submitting, timeout? ${timeout}`);
```
virtuacoplenny: It might be better to make the method simpler by not passing in a boolean and instead have separate logger statements at the call site of submitFeedback. I think it might be better because that removes conditional functionality of submitFeedback. I also think it doesn't really make much a difference but I'm gonna ramble on anyway.

spot-client/src/spot-remote/ui/views/remote-views/feedback.js
```diff
@@ -5,6 +5,13 @@ import { Clock } from 'common/ui';
 
 import { FeedbackForm } from './../../components';
 
+/**
+ * By default will use 1 minute for feedback inactivity timeout.
+ *
+ * @type {number}
+ */
+const DEFAULT_FEEDBACK_TIMEOUT = 60 * 1000;
```
virtuacoplenny: Maybe this should be a defaultProp of FeedbackForm. What do you think?

PR #188, opened 2019-03-22
fix(vendor): init lib-jitsi-meet only for webrtc calls 

virtuacoplenny:
Spot mainly uses lib-jitsi-meet for its MUC handling
and its ProxyPeerConnection. While Spot-TV uses
some of its WebRTC logic, Spot-Remotes does not.
Spot-Remote may be loaded on environments that do
not have full WebRTC support, such as old android
browsers, and preventing app load due to a lack
of WebRTC is unnecessary.

### DIFF
--- a/spot-client/src/common/vendor/jitsi-meet-js-provider.js
+++ b/spot-client/src/common/vendor/jitsi-meet-js-provider.js
@@ -1,35 +1,15 @@
 /* global JitsiMeetJS */
 
 /**
+ * Prevent JitsiMeetJS from spamming the console.
+ */
+JitsiMeetJS.setLogLevel('error');
+
+/**
  * A wrapper around the global JitsiMeetJS, as loaded by lib-jitsi-meet.
  */
 export default {
     get() {
-        if (this._jitsiMeetJS) {
-            return this._jitsiMeetJS;
-        }
-
-        JitsiMeetJS.init({});
-        JitsiMeetJS.setLogLevel('error');
-
-        this._jitsiMeetJS = window.JitsiMeetJS;
-
-        return this._jitsiMeetJS;
-    },
-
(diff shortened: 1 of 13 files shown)


spot-client/src/common/vendor/jitsi-meet-js-provider.js
```diff
@@ -9,7 +11,12 @@ export default {
             return this._jitsiMeetJS;
         }
 
-        JitsiMeetJS.init({});
+        try {
+            JitsiMeetJS.init({});
+        } catch (error) {
+            logger.error('Failed to initialize JitsiMeetJS', { error });
```
paweldomas: I'm thinking it should be left undefined after it fails to initialize. Or is it still used later on ?

paweldomas: If it's used later on then it's a bit weird, because right now we probably assume that we know what went wrong, but that might not be the case in the future and may be causing unexpected problems.

virtuacoplenny: Spot-TV needs the init for gUM related actions. Spot-Remote does not. Both need it for XMPP connection. Maybe the better thing would be to no longer call get() and instead wrap all JitsiMeetJS calls to avoid exposing the global entirely. What do you think?

paweldomas: yes, maybe that will be better

paweldomas: also I was thinking about extract xmpp related stuff to another project, but that's more work

virtuacoplenny:
Force pushed to squash and redo commit message.

PR #192, opened 2019-03-25
ref(screenshare): move device list change listener to av utils

virtuacoplenny:


virtuacoplenny:
Relies on https://github.com/jitsi/lib-jitsi-meet/pull/883. I also want to test with the real screenshare dongle, which I have not done so yet; I wanted to make a draft PR while I wait to test the real dongle but accidentally did not.

PR #193, opened 2019-03-27
Pending states for nav buttons

virtuacoplenny:


virtuacoplenny:
I guess some icons got changed in the design and I didn't notice. I'll have to fix that.

virtuacoplenny:
Force pushed with removal of the wait cursor on buttons in a loading state.

spot-client/src/spot-remote/ui/components/nav/buttons/audio-mute-button.js
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Selects parts of the Redux state to pass in with the props of
+ * {@code AudioMuteButton}.
+ *
+ * @param {Object} state - The Redux state.
+ * @private
+ * @returns {Object}
+ */
+function mapStateToProps(state) {
+    const optimisticMuteState = getOptimisticAudioMuteState(state);
```
paweldomas: I would probably move the logic about "optimistic state" to inside of the selector. If you'd want to display audio mute state anywhere else in the app this logic would have to be duplicated.

PR #194, opened 2019-03-27
fix(nav): change flex stylings for safari not rendering

virtuacoplenny:
On safari, the nav icons would not display the circle border
or the label. Somehow the nested flex wrappers and the
flex direction were causing problems. Admittedly the
problem is not understood but removing stylings fixes
it.


virtuacoplenny:
Image of bug. 
Screen Shot 2019-03-27 at 3 10 12 PM

PR #200, opened 2019-03-31
feat(share): dedicated wireless screenshare flow

virtuacoplenny:


spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -79,3 +104,61 @@ export function setVideoMute(mute) {
...
+    return dispatch => {
+        dispatch(setRequestState(
+            requestTypes.SCREENSHARE,
+            requestStates.PENDING,
+            'proxy'
+        ));
+
+        return remoteControlService.setWirelessScreensharing(true, {
+            onClose: () => dispatch(setLocalWirelessScreensharing(false))
+        })
```
paweldomas: is the indentation of `})` correct here ?

spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -79,3 +104,61 @@ export function setVideoMute(mute) {
...
+            .then(() => {
+                dispatch(setLocalWirelessScreensharing(true));
+                dispatch(setSpotTVState({ screensharingType: 'proxy' }));
+                dispatch(setRequestState(
```
paweldomas: It feels like the logic for doing the request state could be abstracted and re-used across audio/video muted and the screensharing (I still need to check how the code looks there as I don't remember). It could be taking a promise, setting PENDING when called, DONE on resolve and ERROR on reject.

spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -79,3 +104,61 @@ export function setVideoMute(mute) {
...
+                requestTypes.SCREENSHARE, requestStates.ERROR)));
+    };
+}
+
+/**
+ * Stops any screenshare in progress on the Spot-TV.
+ *
+ * @returns {Function}
+ */
+export function stopScreenshare() {
```
paweldomas: what happens when `stopScreenshare()` is called while `remoteControlService.setWirelessScreensharing` called above is still working ? Is that even possible ? Will the start promise get rejected ? Im wondering what will happen to the request state

spot-client/src/spot-remote/ui/views/join-code-entry.js
```diff
@@ -76,31 +78,13 @@ export class JoinCodeEntry extends React.Component {
...
     }
 
-    /**
-     * Redirects to the remote control view when a connection to Spot is
-     * established.
-     *
-     * @inheritdoc
-     */
-    componentDidUpdate(prevProps) {
-        if (!prevProps.isConnectedToSpot && this.props.isConnectedToSpot) {
```
paweldomas: what happened to this logic ? is it no longer relevant ?

spot-client/src/spot-remote/ui/views/share/share.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    /**
+     * Starts the wireless screensharing flow.
+     *
+     * @returns {void}
+     */
+    _onStartWirelessScreenshare() {
+        if (this.props.isScreenshareActiveRemotely) {
+            this.setState({ autoPromptScreenshare: false });
```
paweldomas: What happens when Spot Remote disconnects ? What's the flow in that situation and will the user get auto prompted again ?

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -278,6 +279,11 @@ export class MeetingFrame extends React.Component {
         switch (type) {
         case COMMANDS.HANG_UP:
             this._jitsiApi.executeCommand('hangup');
+
+            if (data.force) {
```
paweldomas: what is the `force` option needed for ? Maybe it would be worth putting a comment wherever that force argument is specified (not here but in the place where it's called).

spot-client/src/spot-remote/ui/views/join-code-entry.js
virtuacoplenny: Yeah, while testing it looked dead and duplicated in another method in this file. I'll double check.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: I like it here and the linter didn't complain. Should it be on line 122?

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
virtuacoplenny: I used force to skip the feedback window. In the share flow, there was no talk of having feedback entry. So for now I skip feedback.

spot-client/src/spot-remote/ui/views/share/share.js
virtuacoplenny: If spot remote disconnects, componentDidUpdate should kick in and take the remote back to join code entry. If the user enters a code again, they would get prompted again yes, because triggering of the desktop picker is controlled by state local to the share view.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: Yes, I agree. I wasn't sure if I should do it for this changeset, because another change needed is having other calls to start/stop screenshare use these actions. Do you feel it's worthwhile to make at least the action-factory change now? I will do it now or soon, so I'm okay with doing it now.

spot-client/src/common/app-state/remote-control-service/actions.js
paweldomas: It feels detached, but I had troubles myself what to do in similar situation. You can leave it there if the linter doesn't complain.

spot-client/src/common/app-state/remote-control-service/actions.js
paweldomas: It can come in a follow up PR.

spot-client/src/spot-remote/ui/views/share/share.js
paweldomas: sounds good

spot-client/src/spot-remote/ui/views/join-code-entry.js
virtuacoplenny: Yeah, this block doesn't get fired on the normal case. I'm not too sure when this block was needed, but right now on join code submit there is already a redirect that happens to  ROUTES.REMOTE_CONTROL or now ROUTES.SHARE. I'm assuming when join code validation was mocked out, and the redirect logic added to the validation, that this block become unnecessary.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: I tried moving it. So now it's:
```
        return remoteControlService.setWirelessScreensharing(
            true,
            { onClose: () => dispatch(setLocalWirelessScreensharing(false)) }
        )
            .then(() => {
```
Maybe the .then shouldn't be indented? Or maybe it should be placed on the same line as the closing parenthesis.

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
virtuacoplenny: Left some new comments. Let me know if there are more changes to be made. Maybe instead of force it should explicitly be skip feedback and renamed later if it does more than skipping feedback.

paweldomas: Yeah, explicit 'skipFeedback' would be nice.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: Let me know if it needs more tweaking.

spot-client/src/common/app-state/remote-control-service/actions.js
paweldomas: is that a bug ?

virtuacoplenny: Yeah

virtuacoplenny: Hid the UI while the work is still being done.

virtuacoplenny:
Force pushed tor resolve conflicts with the default config and config selectors.

PR #202, opened 2019-04-02
feat(setup): save selected devices

virtuacoplenny:


paweldomas:
Actually it seems to be stored and retained even after reloads, so the note should be removed I guess.

virtuacoplenny:
Yes, the note should be removed.

PR #203, opened 2019-04-03
Adds backend prototype

paweldomas:
Check `spot-admin/docs/getting-started.md` to see how to use the backend prototype.

It is optional and the app should still be working without it. It doesn't serve any real purpose yet, but adds new flows for getting the join code and MUC room name.

### DIFF
--- a/spot-client/src/common/backend/utils.js
+++ b/spot-client/src/common/backend/utils.js
@@ -1,0 +1,77 @@
+import { getDeviceId } from 'common/utils/device-id';
+
+/**
+ * FIXME.
+ *
+ * @param {string} serviceEndpointUrl - FIXME.
+ * @returns {Promise<string>}
+ */
+export function fetchJoinCode(serviceEndpointUrl) {
+    const requestOptions = {
+        headers: {
+            'content-type': 'application/json; charset=UTF-8'
+        },
+        body: JSON.stringify({
+            deviceId: getDeviceId()
+        }),
+        method: 'POST',
+        mode: 'cors'
+    };
+
+    return fetch(serviceEndpointUrl, requestOptions)
+        .then(response => {
+            if (!response.ok) {
+                throw Error(
+                    'Failed to fetch the join code:'
+                     + `${response.statusText}, HTTP code: ${response.status}`);
+            }
+
+            return response.json();
+        })
+        .then(json => {
+            if (!json.joinCode) {
+                throw new Error(`No 'joinCode' in the response: ${JSON.stringify(json)}`);
+            }
+
+            return json.joinCode;
+        });
+}
+
+/**
+ * FIXME.
+ *
+ * @param {string} serviceEndpointUrl - FIXME.
+ * @param {string} joinCode - FIXME.
+ * @returns {Promise<Object>}
+ */
+export function fetchRoomInfo(serviceEndpointUrl, joinCode) {
+    if (!joinCode) {
+        return Promise.reject('The \'joinCode\' argument is required');
+    }
+
+    const requestOptions = {
+        method: 'GET',
+        mode: 'cors'
+    };
+
+    return fetch(`${serviceEndpointUrl}?joinCode=${joinCode}`, requestOptions)
+        .then(response => {
+            // FIXME error handling is duplicated with the 'fetchJoinCode'
+            if (!response.ok) {
+                throw Error(
+                    'Failed to fetch the join code:'
+                    + `${response.statusText}, HTTP code: ${response.status}`);
+            }
+
+            return response.json();
+        })
+        .then(json => {
+            if (!json.roomName) {
+                throw new Error(`No 'roomName' in the response: ${JSON.stringify(json)}`);
+            }
+
+            return {
+                roomName: json.roomName
+            };
+        });
+}
--- a/spot-client/src/common/backend/utils.js
+++ b/spot-client/src/common/backend/utils.js
@@ -1,9 +1,84 @@
+import { logger } from 'common/logger';
 import { getDeviceId } from 'common/utils/device-id';
 
 /**
- * FIXME.
+ * FIXME duplicated with post-to-endpoint and the jitter calculation in the loaders.
  *
- * @param {string} serviceEndpointUrl - FIXME.
+ * Gets next timeout using the full jitter pattern.
+ *
+ * @param {number} retry - The retry number. It's 1 on the first retry.
+ * @returns {number} - The amount of waiting before trying another time given in milliseconds.
+ * @private
+ */
+function _getNextTimeout(retry) {
+    // 1st retry 0 - 2 seconds
+    // 2nd retry 0 - 4 seconds
+    // 3rd retry 0 - 16 seconds
+    return Math.floor(Math.random() * Math.pow(2, retry) * 1000);
+}
+
+/**
+ * Sends HTTP request using {@code fetch} with retries on network and server errors.
+ *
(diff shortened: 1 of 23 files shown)


spot-client/src/common/backend/utils.js
```diff
@@ -0,0 +1,146 @@
...
+ * the retry related log entries.
+ * @param {number} retry - Keeps track of the retry attempt.
+ * @param {number} maxRetries - How many times will retry the request.
+ * @returns {Promise<Object>} - A promise resolved with the JSON parsed from the response.
+ */
+function fetchWithRetry(fetchOptions, retry = 0, maxRetries = 3) {
+    const { url, requestOptions, operationName } = fetchOptions;
+
+    return new Promise((resolve, reject) => {
+        fetch(url, requestOptions)
```
virtuacoplenny: I think fetch itself returns a promise so you dont have to wrap it in a promise.

spot-client/src/common/backend/utils.js
```diff
@@ -0,0 +1,146 @@
...
+ * @param {Object} fetchOptions - Options related to the fetch request.
+ * @param {string} fetchOptions.url - The URL to be passed to the fetch function.
+ * @param {Object} fetchOptions.requestOptions - The options to be passed to the fetch function.
+ * @param {string} fetchOptions.operationName - The name of the fetch operation that will appear in
+ * the retry related log entries.
+ * @param {number} retry - Keeps track of the retry attempt.
+ * @param {number} maxRetries - How many times will retry the request.
+ * @returns {Promise<Object>} - A promise resolved with the JSON parsed from the response.
+ */
+function fetchWithRetry(fetchOptions, retry = 0, maxRetries = 3) {
```
virtuacoplenny: I think we can try to do something fancy and create a new function within the scope to avoid having retry and max retry as parameters to the signature. It's not a big deal but maybe later we can try it and maybe you'll love it.

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -434,6 +434,6 @@ export default class XmppConnection {
      * @returns {void}
      */
     _sendPresence() {
-        this.room.sendPresence();
+        this.room && this.room.sendPresence();
```
virtuacoplenny: Were you hitting errors that needed these truthy checks? I can see there being problems in general but am curious what you were seeing.

spot-client/src/common/ui/loaders/abstract-loader.js
```diff
@@ -18,10 +19,45 @@ export class AbstractLoader extends React.PureComponent {
      *
      * @param {Object} props - The read-only properties with which the new
      * instance is to be initialized.
+     * @param {string} serviceName - The name of the service that will be loaded and which will
+     * appear in the logs.
+     * @param {boolean} supportsReconnects=false - Whether or not the service to be loaded by
+     * the {@code AbstractLoader} subclass supports or is willing to do reconnects.
      */
-    constructor(props) {
+    constructor(props, serviceName = 'abstract-loader', supportsReconnects = false) {
```
virtuacoplenny: What's your opinion on calendar and ultrasound also extending abstract loader? Do you think they still should? It's something I can take on to move them off of abstract loader if needed.

spot-client/src/common/utils/device-id.js
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * Obtains the device ID. It's generated randomly on the first run and the retrieved from storage.
+ * That means the ID is persisted between the browser sessions.
+ *
+ * @returns {string}
+ */
+export function getDeviceId() {
+    const deviceId = persistence.get('deviceId') || generateGuid();
+
+    persistence.set('deviceId', deviceId);
```
virtuacoplenny: I'm not sure how I feel about a "get" doing a set.

spot-client/src/spot-remote/ui/loaders/withRemoteControl.js
```diff
@@ -209,50 +175,18 @@ export class RemoteControlLoader extends AbstractLoader {
...
-
-        if (this._unmounted) {
-            logger.warn('Cancelling reconnect due to unmount');
-
-            return Promise.reject();
-        }
-
-        this._isReconnectQueued = true;
-
-        // wait a little bit to retry to avoid a stampeding herd
```
virtuacoplenny: I surprised myself that I account for this. I'm really proud of myself and I'm letting you know that I am.

spot-client/src/common/backend/utils.js
paweldomas: so then how to I `reject` or `resolve` it ?

spot-client/src/common/backend/utils.js
paweldomas: I don't know how to do it. How would I remove the `retry` argument ?

spot-client/src/common/ui/loaders/abstract-loader.js
paweldomas: Yeah, I think they should. At least for now

spot-client/src/common/utils/device-id.js
paweldomas: Me neither... for the outside world it's a get.

spot-client/src/common/remote-control/xmpp-connection.js
paweldomas: Yes, I need to double check when exactly this was happening. I think it's when the joinCode is set and the room is not ready yet.

spot-client/src/common/backend/utils.js
virtuacoplenny: return Promise.reject() or throw an error. resolves should just happen if execution finishes without error.

spot-client/src/common/backend/utils.js
virtuacoplenny: We make a new function within this function that wraps all the logic and either checks retry/maxRetries in its scope or accepts it in the signature. I can give it a shot in an exploratory follow-up PR.

spot-client/src/common/backend/utils.js
paweldomas: How would this part work ?
```
setTimeout(() => {
    fetchWithRetry(fetchOptions, _retry).then(resolve, reject);
}, timeout);
```

spot-client/src/common/backend/utils.js
paweldomas: okay, sounds good

spot-client/src/common/backend/utils.js
virtuacoplenny: Wrap only what needs to be wrapped in a promise instead of all the logic.
```
return new Promise((resolve, reject) => {
  setTimeout(() => {
      fetchWithRetry(fetchOptions, _retry).then(resolve, reject);
  }, timeout);
});

paweldomas: I don't know... somehow for me one outer promise is easier to follow

virtuacoplenny: The rest of the blocks don't need to know about resolve being available in the scope; creating the promise around the timeout would limit the exposure of a variable to only where it's needed. You also don't have to be aware of the outer promise scope while reading through the chain and you won't have to keep context of where the resolve came from and if anything else uses it. Probably more to the point is it'd be easier to break out individual parts of the promise chain into separate replaceable, independent, testable pieces. This is why my preference would be to wrap only the setTimout in a promise. That said, I still approved the PR because I don't think it's important.

PR #205, opened 2019-04-04
ref(backend): use closure to keep track of retry count

virtuacoplenny:


paweldomas:
You can also do this change about not having the outer promise (if you want), or add it as separate commit just to see how will that look

PR #206, opened 2019-04-04
ref(rcs): use a helper for creating spot-command actions

virtuacoplenny:


spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -5,6 +5,41 @@ import { remoteControlService } from 'common/remote-control';
...
+        expectedValue
+    ));
+
+    return request()
+        .then(() => dispatch(setRequestState(
+            requestType,
+            requestStates.DONE,
+            expectedValue
+        )))
+        .then(() => dispatch(setRequestState(requestType, requestStates.ERROR)));
```
paweldomas: did you mean catch here ? also why not use `then(resolved, rejected)` ? doing `.catch` on `then` will also catch errors in `then` which not sure if really intended

virtuacoplenny: Oh wow, yes. catch

virtuacoplenny:
Force pushed to rebase and fix conflicts. In the process squashed the commit to change .then to .catch.

PR #207, opened 2019-04-05
ref(backend): wrap retry timeout in promise

virtuacoplenny:
For consideration.


paweldomas:
Hmm nope, I liked the other version better :P

PR #209, opened 2019-04-06
fix(home): show error message for failed event fetch

virtuacoplenny:


spot-client/src/spot-tv/calendars/outlook.js
```diff
@@ -43,11 +44,25 @@ export default {
...
         const filter = `Start/DateTime ge '${now.toISOString()}'`;
-        const url = `/users/${email}/calendar/events?$filter=${filter}&$top=3`;
+        const url = `/users/${email}/calendfar/events?$filtefdr=${filter}&$top=3`;
         const orderBy = 'createdDateTime ASC';
 
         return microsoftClientApi.request(url, { orderBy })
             .then(response => response.value)
+            .catch(response => {
+                const formattedError = {
+                    errorCode: response.code,
```
paweldomas: `response.message` that comes from Outlook ? It's not part of the Fetch Response API: https://developer.mozilla.org/en-US/docs/Web/API/Response#Properties

spot-client/src/spot-tv/calendars/google.js
```diff
@@ -61,6 +62,20 @@ export default {
 
         return gapi.client.request(calendarEventsEndpoint)
             .then(response => response.result.items)
+            .catch(response => {
+                const formattedError = {
+                    errorCode: response.result.error.errors[0].domain,
```
paweldomas: are you sure `response.result.error` will always be available ? what when it fails on the network layer and there's no response ?

paweldomas: oh it's actually google api and not fetch

PR #210, opened 2019-04-08
feat(feedback): add skip button

virtuacoplenny:
Temporary solution to allow users to skip feedback entry.


virtuacoplenny:
Screen Shot 2019-04-08 at 1 56 13 PM

PR #216, opened 2019-04-10
Calendar backend

paweldomas:


spot-client/src/common/app-state/setup/selectors.js
```diff
@@ -20,6 +20,16 @@ export function getDisplayName(state) {
     return state.setup.displayName;
 }
 
+/**
+ * Returns the JWT token for authentication with backend services.
```
virtuacoplenny: I'm a smug jerk who likes pointing this out: JWT token -> JWT

spot-client/src/common/backend/utils.js
```diff
@@ -85,13 +70,66 @@ function fetchWithRetry(fetchOptions, maxRetries = 3) {
...
+ * @returns {Promise<Array<RESTBackendCalendarEvent>>}
+ */
+export function fetchCalendarEvents(serviceEndpointUrl, jwtToken) {
+    const requestOptions = {
+        method: 'GET',
+        mode: 'cors'
+    };
+
+    if (jwtToken) {
+        requestOptions.headers = new Headers();
```
virtuacoplenny: You don't want to use the constructor options to set authorization and accept?

spot-client/src/common/backend/utils.js
```diff
@@ -85,13 +70,66 @@ function fetchWithRetry(fetchOptions, maxRetries = 3) {
...
+        requestOptions.headers.append('Authorization', `Bearer ${jwtToken}`);
+        requestOptions.headers.append('Accept', 'application/json');
+    }
+
+    let url = serviceEndpointUrl;
+
+    if (!url.includes('{tzid}')) {
+        return Promise.reject(`Missing {tzid} template in the URL: ${url}`);
```
virtuacoplenny: I think early exit error handling should be done at the top if possible. That way the reader doesn't have to follow any other logic that gets thrown away.

spot-client/src/common/backend/utils.js
```diff
@@ -113,7 +151,10 @@ export function fetchJoinCode(serviceEndpointUrl) {
                 throw new Error(`No 'joinCode' in the response: ${JSON.stringify(json)}`);
             }
 
-            return json.joinCode;
+            return {
+                joinCode: json.joinCode,
+                jwt: json.jwt
+            };
```
virtuacoplenny: Is it known how updates to the join code will be pulled/pushed to Spot-TV?

spot-client/src/spot-tv/calendars/backend-calendar.js
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    /**
+     * Requests current calendar events for a provided room.
+     *
+     * @returns {Promise<Array<Object>>}
+     */
+    getCalendar({ jwtToken }) {
+
+        logger.log(`Will fetch calendar events from: ${this.url}`);
```
virtuacoplenny: I think we should use the logger format where the second arg is an object so it's easier to query for.

spot-client/src/spot-tv/calendars/backend-calendar.js
```diff
@@ -0,0 +1,90 @@
...
+        return fetchCalendarEvents(this.url, jwtToken)
+            .then(response => {
+                const events = response && response.events;
+
+                if (!events) {
+                    return Promise.reject('No \'events\' in the response');
+                }
+
+                // Converts to Spot format
+                // FIXME define a type for Spot's calendar event object
```
virtuacoplenny: That would be dope.

spot-client/src/common/app-state/config/default-config.js
```diff
@@ -16,6 +16,9 @@ export default {
         },
         OUTLOOK: {
             CLIENT_ID: process.env.OUTLOOK_CLIENT_ID || ''
+        },
+        BACKEND: {
+            SERVICE_URL: process.env.CALENDAR_SERVICE_URL || ''
```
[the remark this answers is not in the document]
paweldomas: JSON ? REST (not sure if it's rest exactly) ?

spot-client/src/common/app-state/setup/selectors.js
paweldomas: do you want me to also rename all occurrences of `jwtToken` to just `jwt` ?

spot-client/src/common/backend/utils.js
paweldomas: not yet

spot-client/src/common/app-state/setup/selectors.js
virtuacoplenny: Now that you've pointed it out, yes.

spot-client/src/common/app-state/calendars/selectors.js
```diff
@@ -1,3 +1,5 @@
+import { getCalendarConfig } from '../config/selectors';
+import { integrationTypes } from '../../../spot-tv/calendars';
```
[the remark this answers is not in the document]
paweldomas: Not sure I understand the problem here. Is it that we're trying to avoid importing Spot TV or Spot Remote into the common ?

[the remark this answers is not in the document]
virtuacoplenny: Yeah. common should be importing from itself and node_modules; it is the dumping ground for logic that can be used by spot-tv and spot-remote while keeping spot-tv and spot-remote separate. spot-remote should be importing from itself and node_modules and common. spot-tv should be importing from itself and node_modules and common.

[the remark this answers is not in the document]
virtuacoplenny: So for example one way to make me shut up would be move the calendar integration types constants into common so spot-remote doesn't indirectly import something from spot-tv.

spot-client/src/common/app-state/config/default-config.js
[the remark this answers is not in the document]
virtuacoplenny: Backend is sound better and better. I suppose we can also just change it pretty easily later if we ever want to.

spot-client/src/spot-tv/calendars/backend-calendar.js
paweldomas: I removed this log

spot-client/src/common/app-state/config/default-config.js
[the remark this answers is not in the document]
paweldomas: I'm leaving it backend then (for now ... :P)

spot-client/src/common/app-state/setup/selectors.js
paweldomas: done

spot-admin/backend/register-device.js
```diff
@@ -0,0 +1,50 @@
...
+const registerDeviceFailureRate = process.env.REG_DEVICE_FAILURE_RATE;
+
+console.info('register-device failure rate: ' + registerDeviceFailureRate);
+
+const jwtToken = process.env.JWT_TOKEN;
+
+function registerDeviceHandler(spots, req, res) {
+    const { deviceId } = req.body;
+
+    if (!deviceId) {
```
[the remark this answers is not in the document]
paweldomas: okay, I tried looking into that but to much work for a prototype. maybe will try it when adding another controller

spot-client/src/common/app-state/calendars/selectors.js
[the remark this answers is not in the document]
paweldomas: So I moved it to common and then renamed, because it was no longer in "calendar" package. Let me know if that's okay or should I move it to a dedicated "calendar" space in the common.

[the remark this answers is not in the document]
virtuacoplenny: Good enough to me.

paweldomas:
I'm not squashing if that's ok, because it actually wont be so easy with the renames

virtuacoplenny:
> I'm not squashing if that's ok, because it actually wont be so easy with the renames

Seems fine to me; you made clean commit names so I assumed you weren't going to squash, although I don't why the renaming would make squashing trickier.

PR #218, opened 2019-04-11
Add amplitude support

virtuacoplenny:


hristoterezov:
I'm wondering is it going to be possible and does it make sense to extract the analytics implementation and the handlers from jitsi-meet and lib-jitsi-meet and move it to js-utils in order to be able to use it in all js projects? WDYT?

I guess this would be an effort out of the scope of this PR and will require a lot of work but I was wondering have you thought of it?

virtuacoplenny:
> I'm wondering is it going to be possible and does it make sense to extract the analytics implementation and the handlers from jitsi-meet and lib-jitsi-meet and move it to js-utils in order to be able to use it in all js projects? WDYT?

There is so little logic here once jsdocs are removed.

saghul:
> I'm wondering is it going to be possible and does it make sense to extract the analytics implementation and the handlers from jitsi-meet and lib-jitsi-meet and move it to js-utils in order to be able to use it in all js projects? WDYT?
> 
> I guess this would be an effort out of the scope of this PR and will require a lot of work but I was wondering have you thought of it?

Nobody invited me here, but here I go :-) IMHO, a little duplication is much better than a little dependency. Also, js-utils is a weird amalgametion of randrom crap, if we really want to we could have a jitsi-meet-analytics thing instead, just with the 2 (for now) analytics handlers, perhaps.

PR #219, opened 2019-04-11
feat(analytics): add events

virtuacoplenny:


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -137,12 +138,14 @@ export default class DialPad extends React.Component {
     _onGoToCall(event) {
         event.preventDefault();
 
+        analytics.log(meetingJoinEvents.DIAL_OUT);
```
paweldomas: Do you think it makes sense to create actions and add the analytics part there ?

paweldomas:
In general something is telling me it would be better (if possible and makes sense) to have actions for all those things and post analytics from there.

PR #221, opened 2019-04-11
feat(setup): initial styling for device selection

virtuacoplenny:
Missing selector stylings because I'm not sure yet if I want to make components for them or if I can get away with trying to style the select element.

Spot


paweldomas:
When looking at the screenshot, is there an easy way to end the device labels with "..." instead of cutting them off ?

virtuacoplenny:
No, it's not that easy which is why this PR is missing selector styling. I'm unsure if I want to bring in a dropdown component, make my own, or continue trying to re-style the HTML select element.

PR #223, opened 2019-04-12
fix(spot-tv): consistent spot remote url display

virtuacoplenny:


spot-client/src/common/utils/window-handler.js
```diff
@@ -10,7 +10,22 @@ export default {
      * @returns {string}
      */
     getBaseUrl() {
-        return window.location.origin;
+        const origin = window.location.origin;
```
paweldomas: is this code changing on what the function returns ? are the JSDocs still valid here ?

virtuacoplenny: Sort of. I actually think they were never really accurate to begin with. I'll tweak.

spot-client/src/spot-tv/ui/views/home.js
```diff
@@ -336,6 +331,16 @@ export class Home extends React.Component {
...
     }
+
+    /**
+     * Generates the URL for a browser to become a Spot-Remote.
+     *
+     * @private
+     * @returns {string}
+     */
+    _getSpotRemoteConnectUrl() {
+        return `${windowHandler.getBaseUrl()}/${this.props.joinCode.toUpperCase()}`;
```
paweldomas: okay, now the logic for generating the join URL is duplicated in 3 places, is there a way to move this into a helper function or something ?

virtuacoplenny: I've moved all the logic into one component. Generally the link creation logic is for debugging and the display of the full URL was never per design. So we'll see what happens when we get more design tweaks.

virtuacoplenny: So it looks like in the future the base part won't even exist. But for now we'll keep it,

PR #229, opened 2019-04-15
Add analytics event for feedback

paweldomas:
This commit can probably be split into two parts where one is creating action for submitting the feedback and another one which will add the analytics event.

### DIFF
--- a/spot-client/src/spot-remote/analytics/index.js
+++ b/spot-client/src/spot-remote/analytics/index.js
@@ -1,0 +1,1 @@
+import './middleware';
--- a/spot-client/src/spot-remote/remote-control/middleware.js
+++ b/spot-client/src/spot-remote/remote-control/middleware.js
@@ -1,0 +1,20 @@
+import { MiddlewareRegistry } from 'common/redux';
+import { remoteControlService } from 'common/remote-control';
+
+import { SUBMIT_FEEDBACK } from './action-types';
+
+MiddlewareRegistry.register(() => next => action => {
+    const result = next(action);
+
+    switch (action.type) {
+    case SUBMIT_FEEDBACK: {
+        remoteControlService.submitFeedback({
+            score: action.score,
+            message: action.message
+        });
+        break;
+    }
+    }
+
+    return result;
+});
--- a/spot-client/src/common/analytics/events.js
+++ b/spot-client/src/common/analytics/events.js
@@ -1,0 +1,4 @@
+export const feedbackEvents = {
+    SKIP: 'feedback-skip',
+    SUBMIT: 'feedback-submit'
+};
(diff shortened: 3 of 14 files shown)


spot-client/src/spot-remote/analytics/index.js
```diff
@@ -0,0 +1 @@
+import './middleware';
```
virtuacoplenny: I'm actually not a fan of an import causing side effects. I'd much prefer, stylistically, to have something import this and apply it explicitly. I can't justify completely how or why I feel this way, but it's a bit less surprising to me to have to manually initialize. What is your opinion?

spot-client/src/spot-remote/remote-control/middleware.js
```diff
@@ -0,0 +1,20 @@
...
+
+    switch (action.type) {
+    case SUBMIT_FEEDBACK: {
+        remoteControlService.submitFeedback({
+            score: action.score,
+            message: action.message
+        });
+        break;
+    }
+    }
```
virtuacoplenny: So the next step after all this work settles would be  bring (maybe) all rcs logic calls into here?

spot-client/src/spot-remote/analytics/index.js
virtuacoplenny: But just now I remembered how we’re now in a mode of moving everything into middleware: because I tried to be different from jitsi-meet. So nevermind, let’s do whatever jitsi meet does to make choices easier.

paweldomas: yeah I agree, but I just don't know how to do any better without extra manual work. It seems that once the feature is imported it kind of enables the middleware.

spot-client/src/spot-remote/remote-control/middleware.js
paweldomas: yeah, I'm thinking all calls done by the UI should be converted to be actions.

spot-client/src/spot-remote/analytics/index.js
paweldomas: well, maybe I will come up with something eventually, as I'm trying to enable it only for Spot Remote and it was done by importing the module here:
https://github.com/jitsi/jitsi-meet-spot/blob/add-analytics-events2/spot-client/src/spot-remote/ui/views/remote-control.js#L15
Which is probably not good enough, because we will probably need the analytics for the `join-code-entry` as well

paweldomas: I don't know it doesn't seem to be very urgent. Anyway I'm wondering if it would be okay to import the analytics again from `join-code-entry` until there is that bundle ? Seems like the easiest thing to do really

virtuacoplenny: Importing twice would not register the middleware twice.

paweldomas: then I really kind of like it as it seems there's a bit of automagic in this approach as well

PR #233, opened 2019-04-16
ref: replace inputs with material input

virtuacoplenny:
Ignores inputs where temporary styling is used.


virtuacoplenny:
Looks like I broke tests.

PR #239, opened 2019-04-19
More analytics

paweldomas:


spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -92,7 +182,16 @@ export function goToMeeting(meetingName, options) {
...
-    return () => remoteControlService.hangUp(skipFeedback);
+    return dispatch => {
+        remoteControlService
+            .hangUp(skipFeedback)
+            .then(() => {
+                dispatch({
+                    type: HANG_UP,
+                    skipFeedback
+                });
+            });
```
virtuacoplenny: Why do this dispatch in the promise success but in the other actions the dispatch happens outside the promise chain?

spot-client/src/spot-remote/analytics/middleware.js
```diff
@@ -15,6 +52,31 @@ MiddlewareRegistry.register(() => next => action => {
...
+
+    switch (requestType) {
+    case requestTypes.AUDIO_MUTE:
+        analytics.log(inCallEvents.AUDIO_MUTE, { muting: expectedState });
+        break;
+    case requestTypes.VIDEO_MUTE:
+        analytics.log(inCallEvents.VIDEO_MUTE, { muting: expectedState });
+        break;
+    }
+}
```
virtuacoplenny: Looking at amplitude the user timeline looks clear enough to you? I'm a little worried about the timeline getting too noisy but I'll trust your judgement.

paweldomas: I think it's okay so far. I'm really trying to implement events from your other PR, so Im thinking you were kind of okay with logging audio and video mute.

virtuacoplenny: My comment shouldn’t have been done on a specific line. (It’s a habit of mine to leave the comment whenever I think of it.) I was asking in general and not about mute specifically, but the answer would still be it looks okay.

PR #241, opened 2019-04-22
ref(rcs): move connection logic into middleware

virtuacoplenny:
- Shove all connection creation logic into middleware.
  This is done to match jitsi-meet's structure to
  make it easier for people to onboard into.
- Create a bootstrap event for middleware to hook onto
  to know when to load RCS. This mirrors jitsi-meet's
  APP_WILL_MOUNT.
- Export all the features of app-state for convenience
  and to mirror how jitsi-meet exports all of it.
- Treat connection establishing as a request for
  timing display of loading indicators.
- Change RCS cleanup to immediately remove references
  to any existing connection. This is done to allow
  immediate reconncets, as the destroy command takes
  several seconds for an unknown reason.
- Store an instance of history so middleware can
  access it.
- Create a new "loader" component that shows a
  loading icon for the remote control service connection
  but does not loading on its own.


virtuacoplenny:
I dont get the conflicts so I'm just gonna close.

PR #244, opened 2019-04-23
ref(remote): use actions for connection

virtuacoplenny:


spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -42,11 +50,15 @@ function createActionWithRequestStates( // eslint-disable-line max-params
     ));
 
     return request()
-        .then(() => dispatch(setRequestState(
-            requestType,
-            requestStates.DONE,
-            expectedValue
-        )), error => {
+        .then((...args) => {
```
paweldomas: hmm what happened here ? maybe the docs need to be updated to explain this ?

paweldomas: yeah it would be good to mention that `@returns {Function<Promise>}` is resolved with the result of `request`

spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -59,6 +71,81 @@ function createActionWithRequestStates( // eslint-disable-line max-params
         });
 }
 
+/**
+ * Establishes a connection to an existing Spot-MUC using the provided join code.
+ *
+ * @returns {Object}
+ */
+export function createSpotTVRemoteControlConnection() {
```
paweldomas: I'm thinking maybe this should not be placed in the 'common' package since it's Spot TV specific

virtuacoplenny:
Do you think the action is the right place for this? Do you think something should be added to prevent multiple calls to create a connection?

paweldomas:
I think the action is okay. Something that detects multiple connections being created will probably be useful to prevent unexpected issues.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: I'll have to think about how to move this action and "createActionWithRequestStates".

PR #249, opened 2019-04-23
fire event change event on initial fetch

virtuacoplenny:


spot-client/src/spot-tv/calendars/calendarService.js
```diff
@@ -34,6 +34,8 @@ export class CalendarService extends EventEmitter {
          * A cache of the last calendar events that have been fetched.
          */
         this._calendarEvents = [];
+
+        this._hasFetchedEvents = false;
```
paweldomas: hmm did you mean `_hasUpdatedEvents` or to use `_hasFetchedEvents` in the code below ? the var is named differently in all 3 places

virtuacoplenny: I'll spend the rest of the day trying to forget this ever happened.

PR #250, opened 2019-04-24
ref(redux): generalize actions with request state updates

virtuacoplenny:
So it can be re-used for other actions which create async requests.


spot-client/src/spot-remote/analytics/middleware.js
```diff
@@ -76,43 +92,25 @@ MiddlewareRegistry.register(() => next => action => {
...
-        } else if (expectedState === undefined) {
-            analytics.log(inCallEvents.SCREENSHARE_STOP);
-        }
-        break;
-    case requestTypes.VIDEO_MUTE:
-        analytics.log(inCallEvents.VIDEO_MUTE, { muting: expectedState });
-        break;
-    }
+function isPendingAsyncAction(action) {
+    return action.requestState !== asyncActionRequestStates.PENDING;
```
paweldomas: shouldn't it be `action.requestState === asyncActionRequestStates.PENDING` ?

virtuacoplenny: :(

PR #251, opened 2019-04-24
ref(setup): move wired screenshare selector to device selection

virtuacoplenny:
Screen Shot 2019-04-24 at 12 35 39 PM


paweldomas:
I like that ! Is there an option to select no device for the screensharing ?

virtuacoplenny:
Nope, there is no "none" option. I didn't add it for three main reasons:
1. Not called out by design.
2. Wasn't sure if it'd get lost in the current PR.
3. Overcome with laziness so I want to do it later.

I'll let you decide which of those is the real, real reason why I didn't do it.

PR #254, opened 2019-04-25
feat(setup): new flow for backend

virtuacoplenny:
Use a separate setup flow when there is an admin service url defined. Leave in the old flow so we can keep developing and testing other features.

Screen Shot 2019-04-24 at 6 18 15 PM
Screen Shot 2019-04-24 at 6 18 25 PM
Screen Shot 2019-04-24 at 6 18 28 PM


virtuacoplenny:
Note: the first step of the new flow, pairing with a backend, is a no-op.

paweldomas:
is it possible to go back in the setup flow ?

virtuacoplenny:
It has not been possible to go back in setup flow. setup flow has been under one route, /setup.

PR #264, opened 2019-04-29
Setup flow fixes

paweldomas:
Minor changes before the big ones will come in


virtuacoplenny:
I'm pleasantly surprised everything still works after removing the calendar loader.

PR #270, opened 2019-05-10
feat(meeting): add large audio and video mute icons

virtuacoplenny:
Screen Shot 2019-05-09 at 5 12 41 PM

### DIFF
--- a/spot-client/src/spot-tv/ui/components/meeting-status/meeting-status.js
+++ b/spot-client/src/spot-tv/ui/components/meeting-status/meeting-status.js
@@ -29,16 +29,16 @@ export class MeetingStatus extends React.Component {
             <div className = 'meeting-status'>
                 {
                     this.props.audioMuted && (
-                        <StatusIcon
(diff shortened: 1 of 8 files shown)


spot-client/src/spot-tv/ui/components/meeting-status/meeting-status.js
```diff
@@ -0,0 +1,63 @@
...
+     * @inheritdoc
+     * @returns {ReactElement}
+     */
+    render() {
+        return (
+            <div className = 'meeting-status'>
+                {
+                    this.props.audioMuted && (
+                        <StatusIcon
+                            icon = { <MicOff /> }
```
zbettenbuk: Don't you wanna pass the component class instead of an instantiated component? Not sure what this causes, it just feels weird to me.

virtuacoplenny: Maybe a child instead?

PR #275, opened 2019-05-11
ref(tests): replace screen-info with custom screen count utils

virtuacoplenny:
screen-info is no longer being maintained and has an outdated
package with a security vulnerability.


virtuacoplenny:
Force pushed to fix conflicts.

PR #277, opened 2019-05-11
ref(deps): use dayjs over moment

virtuacoplenny:
Moment is 234kb while dayjs is 7kb, both rounded up
to the nearest kb.


virtuacoplenny:
Force pushed to fix typos in commit message.

PR #279, opened 2019-05-12
fix(code-entry): work around android backspace moving focus forward

virtuacoplenny:
Note: this is entirely a workaround and only a partial patch to the UX on android.


virtuacoplenny:
One workaround Zoli suggested was assuming that any keycode 229 event is a backspace and moving backwards in the code input if there is no input value and the keycode is 229. That sounds okay-ish and maybe something to play around with for a follow up PR.

PR #280, opened 2019-05-13
fix(bootstrap): point jitsi-meet iframe api to prod

virtuacoplenny:


virtuacoplenny:
I don't really know what we want to do here, to be honest. We really shouldn't be pointing it at beta, which can go down whenever. Using meet.jit.si would be the documented way to consume the iframe library, but we will have to wait for deploys to get api changes.

virtuacoplenny:
Chatted with Pawel a bit about this. We agree pointing to beta is bad but spot still isn't technically released so we can live with this for the time being. I'm going to make a ticket so we don't forget this issue and we'll revisit in the future.

PR #283, opened 2019-05-13
fix(tests): update tests for auto-code submission

virtuacoplenny:
Amateur hour.


virtuacoplenny:
I had force pushed to resolve linting errors I caught immediately after creating the PR because it's amateur hour.

PR #284, opened 2019-05-13
Few fixes for running the in-room-controller app in release mode

paweldomas:


in-room-controller/App.js
```diff
@@ -55,11 +55,16 @@ export default class App extends React.Component {
...
+            // The side menu for editing the URL is enabled only in the debug mode.
+            if (__DEV__) {
+                return (
+                    <SideMenu menu = { this._renderSettingsMenu() }>
+                        <RemoteControl url = { remoteControlUrl } />
+                    </SideMenu>
+                );
+            }
+
+            return <RemoteControl url = { remoteControlUrl } />;
```
paweldomas: is there a better way of doing it ? is it possible to wrap into parent component conditionally ?

paweldomas: okay I googled something, will update the PR

virtuacoplenny: process.env?

paweldomas:
Force pushed, because there were eslint errors

in-room-controller/.eslintrc
```diff
@@ -4,6 +4,9 @@
         "eslint-config-jitsi/jsdoc",
         "eslint-config-jitsi/react"
     ],
+    "globals": {
+        "__DEV__": true
```
virtuacoplenny: How would this get set to false?

paweldomas: oh I merged before seeing your reply. So this is only for the eslint. It is set by react-native to false when running in release mode

virtuacoplenny: Ah, right, true meaning it's an allowed global.

PR #285, opened 2019-05-14
Initial commit of spot-electron

zbettenbuk:
Initial commit of the electron app with the loadURL approach as we agreed.

This now supports
- Persistent config
- Window mode changes
- Logging

### DIFF
--- a/spot-electron/.eslintrc
+++ b/spot-electron/.eslintrc
@@ -1,0 +1,16 @@
+{
+    "extends": [
+        "eslint-config-jitsi",
+        "eslint-config-jitsi/jsdoc"
+    ],
+    "rules": {
+        "no-console": 1,
+        "space-before-function-paren": ["error", {
+            "anonymous": "always",
+            "named": "never",
+            "asyncArrow": "always"
+        }]
+    },
+    "settings": {
+    }
+}
--- a/spot-electron/config.js
+++ b/spot-electron/config.js
@@ -1,0 +1,24 @@
+const process = require('process');
+
+module.exports = {
+
+    /**
+     * The default URL to connect to.
+     */
+    defaultSpotURL: process.env.SPOT_URL || 'https://spot.jitsi.net/spot',
+
+    /**
+     * The details of the spash screen, if any.
+     */
+    // spashScreen: {
+    //     height: 192,
+    //     logo: 'static/applogo.png',
+    //     width: 192
+    // },
+
+    window: {
+        fullscreen: undefined,
+        height: 800,
+        width: 1200
+    }
+};
--- a/spot-electron/package.json
+++ b/spot-electron/package.json
@@ -1,0 +1,35 @@
+{
+    "author": "Jitsi",
+    "build": {
+        "appId": "org.jitsi.spot",
+        "mac": {
+            "category": "public.app-category.business",
+            "target": "default"
+        }
+    },
+    "description": "",
+    "devDependencies": {
+        "babel-eslint": "10.0.1",
+        "electron": "5.0.1",
+        "electron-builder": "20.40.2",
+        "eslint": "5.16.0",
+        "eslint-config-jitsi": "github:jitsi/eslint-config-jitsi#1.0.1",
+        "eslint-plugin-flowtype": "3.8.2",
+        "eslint-plugin-import": "2.17.2",
+        "eslint-plugin-jsdoc": "4.8.4"
+    },
+    "license": "ISC",
+    "main": "main.js",
+    "name": "spot-electron",
+    "productName": "Jitsi Spot",
+    "scripts": {
+        "dist": "electron-builder",
+        "lint": "eslint .",
+        "pack": "electron-builder --dir",
+        "start": "electron ."
+    },
+    "version": "1.0.0",
+    "dependencies": {
+        "lodash": "4.17.11"
+    }
+}
--- a/spot-electron/src/application-window/applicationwindow.js
+++ b/spot-electron/src/application-window/applicationwindow.js
@@ -1,0 +1,106 @@
+const {
+    BrowserWindow
+} = require('electron');
+
+const {
+    defaultSpotURL,
+    spashScreen
+} = require('../../config');
+
+const { config } = require('../config');
+
+/**
+ * This is the reference for the main window. May be exposed from this scope later, but we don't have
+ * any reasons to do it yet.
+ */
+let applicationWindow;
+
+/**
+ * Reference to the splash window. This can be disabled in the config.
+ */
+let splashWindow;
+
+/**
+ * Function to create the main window (application window).
+ *
+ * @returns {void}
+ */
+function createApplicationWindow() {
+    applicationWindow = new BrowserWindow({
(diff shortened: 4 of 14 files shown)


spot-electron/src/logger/logger.js
```diff
@@ -0,0 +1,53 @@
...
+     * @param {?Error} throwable - The error to be logged, if any.
+     * @returns {void}
+     */
+    _log(level, message, throwable) {
+        // eslint-disable-next-line no-console
+        console.log(`${new Date().toISOString()} - ${level} - ${message}${throwable ? ` - ${throwable}` : ''}`);
+    }
+}
+
+exports.logger = new Logger();
```
virtuacoplenny: Here are some random musings I have about logging. In the future should we be sending these logs somewhere like spot-client does? Should we store them locally? I think in the future we should also add an uncaught exception handler (I forget how I did that previously). Also, should we be using jitsi-meet-logger here too?

spot-electron/Makefile
```diff
@@ -0,0 +1,3 @@
+default:
+	# Build spot-client and get the artifacts here
+	cd ../spot-client && npm run build -- --env.target=../spot-electron
```
virtuacoplenny: Do you still need this makefile?

spot-electron/config.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
+
+    /**
+     * The default URL to connect to.
+     */
+    defaultSpotURL: 'https://spot.8x8.vc/spot',
```
virtuacoplenny: Maybe add a process.env variable right now to make this configurable immediately? That would be my preference, but not a requirement.

spot-electron/package.json
```diff
@@ -0,0 +1,36 @@
+{
+    "author": "Jitsi",
+    "build": {
+        "appId": "org.jitsi.spot",
+        "mac": {
+            "category": "public.app-category.business",
+            "target": "default"
+        },
+        "productName": "Jitsi Spot"
```
virtuacoplenny: Is it Jitsi Spot or Jitsi-Meet Spot? This repo is called jitsi-meet-spot. We should get verification at some point in the future.

spot-electron/package.json
```diff
@@ -0,0 +1,36 @@
...
+        "eslint": "5.16.0",
+        "eslint-config-jitsi": "github:jitsi/eslint-config-jitsi#1.0.1",
+        "eslint-plugin-flowtype": "3.8.2",
+        "eslint-plugin-import": "2.17.2",
+        "eslint-plugin-jsdoc": "4.8.4"
+    },
+    "license": "ISC",
+    "main": "main.js",
+    "name": "spot-electron",
+    "productName": "Jitsi Spot",
```
virtuacoplenny: Do you need productName here and under build?

spot-electron/src/application-window/applicationwindow.js
```diff
@@ -0,0 +1,106 @@
...
+const {
+    defaultSpotURL,
+    spashScreen
+} = require('../../config');
+
+const { config } = require('../config');
+
+/**
+ * This is the reference for the main window. May be exposed from this scope later, but we don't have
+ * any reasons to di it yet.
```
virtuacoplenny: to *hide it yet?

spot-electron/Makefile
zbettenbuk: Nope, thanks!

spot-electron/config.js
zbettenbuk: Good call, thanks!

spot-electron/package.json
zbettenbuk: Asked Emil, hopefully we get an answer in a few weeks

spot-electron/src/application-window/applicationwindow.js
zbettenbuk: to *do it yet

thanks!

spot-electron/package.json
zbettenbuk: I thought we do, one for the app and one for the packager, but I removed the one under build and it seems like picking up the one on the top level. Thanks!

spot-electron/.eslintrc
```diff
@@ -0,0 +1,16 @@
...
+        "eslint-config-jitsi/jsdoc"
+    ],
+    "rules": {
+        "no-console": 1,
+        "space-before-function-paren": ["error", {
+            "anonymous": "always",
+            "named": "never",
+            "asyncArrow": "always"
+        }]
+    },
```
virtuacoplenny: Why the rule override for space-before-function-paren? I wonder why I did the same for spot-webdriver...

spot-electron/config.js
```diff
@@ -0,0 +1,24 @@
...
+     * The default URL to connect to.
+     */
+    defaultSpotURL: process.env.SPOT_URL || 'https://spot.8x8.vc/spot',
+
+    /**
+     * The details of the spash screen, if any.
+     */
+    spashScreen: {
+        height: 192,
+        logo: 'static/applogo.png',
```
virtuacoplenny: Is this the jitsi logo?

spot-electron/package.json
```diff
@@ -0,0 +1,35 @@
...
+    "scripts": {
+        "dist": "electron-builder",
+        "lint": "eslint .",
+        "pack": "electron-builder --dir",
+        "start": "electron ."
+    },
+    "version": "1.0.0",
```
virtuacoplenny: I wonder if we shouldn't make it v1. I wonder if it makes any difference at all right now.

spot-electron/src/config/config.js
```diff
@@ -0,0 +1,92 @@
+const { app } = require('electron');
+const fs = require('fs');
+const _ = require('lodash');
+const { logger } = require('../logger');
+const defaultConfig = require('../../config');
+
+const PERSISTED_CONFIG = `${app.getPath('userData')}/config.json`;
```
virtuacoplenny: Do you if this type of path accessing with forward slash also work on windows? (not saying change it now if needed)

spot-electron/.eslintrc
zbettenbuk: I reused your file :D

spot-electron/config.js
zbettenbuk: Can be, not now. Yana said we should use this for now

saghul: Me no likey. It doesn't make sense that this is not the Jitsi logo. There is no precedent, even.

zbettenbuk: Removed, as well as replaced the default URL

spot-electron/src/config/config.js
zbettenbuk: Good catch, I should have used node path

spot-electron/package.json
zbettenbuk: I think this is something yet to sort out. I'm okay with any kind of versioning

virtuacoplenny: i just remembered we do date versioning (YYYY.MM.DD) for spot web releases. We should be doing that for mobile as well so probably we'll end up doing that for electron if possible.

PR #286, opened 2019-05-14
ref(code-input): rewrite with one input approach

virtuacoplenny:
Instead of using 6 input elements, use one, which
will allow backspace to work on Android browser.

android

### DIFF
--- a/spot-client/src/common/ui/components/code-input/code-input.js
+++ b/spot-client/src/common/ui/components/code-input/code-input.js
@@ -3,14 +3,11 @@ import React from 'react';
 
 import { isAutoFocusSupported } from 'common/utils';
 
-import InputBox from './input-box';
-
 /**
  * A mapping of keyboard key names to keyboard key codes for readability while
  * processing key codes.
  */
 const keyCodes = {
-    BACKSPACE_KEY: 8,
     LEFT_ARROW_KEY: 37,
     UP_ARROW_KEY: 38,
     RIGHT_ARROW_KEY: 39,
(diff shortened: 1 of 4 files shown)


virtuacoplenny:
There are currently two know issues with this approach:
1. On android browser, when refocusing on the hidden input a cursor will display with a blue circle below it, and that blue circle I cannot figure out how to hide so there can be a blue circle floating on the page.
2. Using one input prevents leveraging mobile browser's auto-centering on the various boxes. So I will have to make another PR, which should be merged before this one, which deviates from designs by making code input boxes display in one row to side step the lack of auto-centering.

virtuacoplenny:
The 6x1 grid PR is here: https://github.com/jitsi/jitsi-meet-spot/pull/288.

virtuacoplenny:
The amount of cross browser issues popping is becoming so large I'm going to mark this as WIP.

spot-client/src/common/ui/components/code-input/code-input.js
```diff
@@ -3,22 +3,19 @@ import React from 'react';
...
 
-import InputBox from './input-box';
-
 /**
  * A mapping of keyboard key names to keyboard key codes for readability while
  * processing key codes.
  */
 const keyCodes = {
-    BACKSPACE_KEY: 8,
     LEFT_ARROW_KEY: 37,
```
zbettenbuk: Do you still need these?

virtuacoplenny: I use them to prevent manual cursor movement on browsers, so the cursor is always at the end of the inputed value.

PR #292, opened 2019-05-15
ref(deps): remove nosleep

virtuacoplenny:
Desired UX is now to let browsers sleep and
rely on reconnect logic.


virtuacoplenny:
Made a commit message typo so force pushed to fix it.

PR #294, opened 2019-05-16
fix(layout): prevent background from resizing with overflow

virtuacoplenny:
Notice the background not change as the page overflows. In the past we wanted to prevent overflow but likely that isn't possible without a lot more design consideration and maybe some special casing (not sure). So let's support overflow a little better by not making the background stretch as the overflow appears and disappears.

May-15-2019 18-52-51


virtuacoplenny:
> Also I saw that we have auto-capitalise ON on the room field on mobile (at least it does it for me). Do you think we should remove that?

Okay I'll turn it off.

PR #295, opened 2019-05-16
Volume control

zbettenbuk:
@virtuacoplenny @paweldomas this is a draft PR to get some feedback if I'm doing everything right doing spot UI. This is quite different from meet, so I'm not sure I understood the structure and the coding style properly. Any feedback are appreciated


virtuacoplenny:
Are you saying it's quite different because it doesn't have a modal manager? Style wise it still uses jitsi-meet's eslint rules.

zbettenbuk:
No, I mean the file/directory structure, the redux handling...etc

virtuacoplenny:
Let me known somewhere else if there are changes you want there. I'm open to doing changing stuff to make it easier for jitsi people to come into the repo (while still trying to avoid circular dependencies)

spot-client/src/common/css/modal.scss
```diff
@@ -14,6 +14,7 @@
 
     .modal-shroud {
         height: 100%;
+        pointer-events: fill;
```
virtuacoplenny: I'm not really a fan of this UX for this change. It looks odd to me that the other buttons on the page look clickable but actually are not.

spot-client/src/spot-remote/ui/components/remote-control-menu/volume-button.js
```diff
@@ -0,0 +1,52 @@
...
+
+const TYPE_DOWN = 'down';
+const TYPE_UP = 'up';
+
+/**
+ * Implements a volume control button.
+ */
+export default class VolumeButton extends React.Component {
+    static propTypes = {
+        type: PropTypes.oneOf(TYPE_DOWN, TYPE_UP)
```
virtuacoplenny: I think there's a propType warning from this. `A common mistake is to write oneOf(x, y, z) instead of oneOf([x, y, z]).`

spot-client/src/spot-remote/ui/components/remote-control-menu/menu-button.js
```diff
@@ -0,0 +1,17 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const MenuButton = ({ label, onClick }) => (
+    <button
```
virtuacoplenny: I do have a nav button that provides similar stylings to the other buttons in the remote-control. Would it be tricky to try to re-use that?

spot-client/src/common/css/button.scss
```diff
@@ -44,4 +44,37 @@
...
+        border-radius: 5px;
+        color: white;
+        font-size: 1em;
+        height: 48px;
+    }
+
+    &.volume {
+        background-color: rgba(0, 0, 0, 0.5);
+        height: 96px;
+        width: 96px;
```
virtuacoplenny: I've generally been using CSS variables for some sizes so that I can scale sizes across different device widths. Do you think it's too much? At some point I started feeling like it was unnecessary overhead.

spot-client/src/common/css/page-layouts/more-modal.scss
```diff
@@ -0,0 +1,9 @@
+@import './../variables.scss';
+
+.more-modal {
+    background-color: var(--container-bg-color);
+    display: flex;
+    flex: 1;
+    flex-direction: column;
+    padding: 50px;
+}
```
virtuacoplenny: We should probably prioritize a modal manager to be done sooner rather than later now. It can probably have all these styles.

spot-client/src/common/css/page-layouts/volume-modal.scss
```diff
@@ -0,0 +1,18 @@
...
+.volume-modal {
+    align-items: center;
+    background-color: var(--container-bg-color);
+    display: flex;
+    flex: 1;
+    flex-direction: column;
+    padding: 50px;
+
+    span {
+        background-color: rgba(0, 0, 0, 0.5);
```
virtuacoplenny: I try, and probably fail, at keeping colors as CSS variables. Sometimes I give up because there are so many one-off colors. I want CSS variables because that way in the future we can do custom color theming using JS, although that plan might have some bumps because Yana had mentioned once using material-UI and using its theming ability.

spot-client/src/common/remote-control/remote-control-service.js
```diff
@@ -208,6 +208,17 @@ class RemoteControlService extends EventEmitter {
             });
     }
 
+    /**
+     * Requests a Spot-TV to change its audio level to a specific direction.
+     *
+     * @param {string} direction - The direction of the volume adjustment. One of 'up' or 'down'.
+     * @returns {Promise} Resolves if the command has been acknowledged.
+     */
+    adjustVolume(direction) {
```
virtuacoplenny: Maybe this should be a number instead of a direction? Like +1, -1? That's not needed now though.

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
```diff
@@ -0,0 +1,45 @@
+import { logger } from 'common/logger';
+
+/**
+ * Implements a controller that utilizes the event emitter of the native (Electron) process if available.
+ *
+ * Using it will enable SpotTVs running wrapped in Electron access OS-native functions.
+ */
+class NativeControntroller {
+    /**
+     * Some.
```
virtuacoplenny: Day. I'll. Wish. Upon. A. Star.

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
```diff
@@ -0,0 +1,45 @@
...
+     */
+    constructor() {
+        // eslint-disable-next-line
+        this.ipcRenderer = this.isAvailable() && __non_webpack_require__('electron').ipcRenderer;
```
virtuacoplenny: Did you still want to change this to window.require?

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    /**
+     * Returns true if the native functions are available (electron env.), false otherwise (web env.).
+     *
+     * @returns {boolean}
+     */
+    isAvailable() {
+        return window && window.process && window.process.emit;
```
virtuacoplenny: I can a silly browser check utility somewhere that this might be better off stuffed in. I should probably re-use it for the electron desktop picker.

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
```diff
@@ -0,0 +1,45 @@
...
+                command,
+                args
+            });
+        } else {
+            logger.log(`Native functions are not available to do '${command}'`);
+        }
+    }
+}
+
+export default new NativeControntroller();
```
virtuacoplenny: What do you think about electron adding a global native controller variable. That way we can do feature detection per feature based on what's on that global. I don't know if that's even still possible to do but it's what was done on a previous project I worked on.

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -290,6 +292,10 @@ export class MeetingFrame extends React.Component {
         });
 
         switch (type) {
+        case COMMANDS.ADJUST_VOLUME:
+            adjustVolume(data.direction);
```
virtuacoplenny: You know, the interesting part to me is this functionality...doesn't have to be meeting specific. That you put the handler in the meeting-frame makes me think "why in meeting only?"

spot-electron/src/application-menu/applicationmenu.js
```diff
@@ -21,4 +21,4 @@ const menuTemplate = [
 
 const menu = Menu.buildFromTemplate(menuTemplate);
 
-Menu.setApplicationMenu(menu);
+// Menu.setApplicationMenu(menu);
```
virtuacoplenny: What is this change for?

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
```diff
@@ -0,0 +1,45 @@
...
+     * @param {string} command - The command identifyer.
+     * @param {Object} args - Args of the command, if any.
+     * @returns {void}
+     */
+    sendMessage(command, args) {
+        if (this.isAvailable()) {
+            this.ipcRenderer.send('native-command', {
+                command,
+                args
+            });
```
virtuacoplenny: Is there a way to get confirmation back that the command finished or succeeded? I don't think it's needed now (and to be fair may never be needed) but it'd be nice to know we can use something like a promise if needed.

spot-client/src/spot-remote/ui/components/remote-control-menu/volume-button.js
```diff
@@ -0,0 +1,52 @@
...
+        );
+    }
+
+    /**
+     * Callback to handle the onClick event of the button.
+     *
+     * @returns {void}
+     */
+    _onClick() {
+        remoteControlService.adjustVolume(this.props.type);
```
virtuacoplenny: We might want to dispatch this in an action. We started talking to remoteControlService (rcs) through actions, and we have an analytics middleware to listen for actions.

spot-client/src/spot-remote/ui/components/remote-control-menu/volume-button.js
```diff
@@ -0,0 +1,52 @@
...
+            </button>
+        );
+    }
+
+    /**
+     * Callback to handle the onClick event of the button.
+     *
+     * @returns {void}
+     */
+    _onClick() {
```
virtuacoplenny: Maybe in the future we need to be fancy and allow long-press/click-hold to go through the volume at larger increments (or do that slider).

spot-electron/src/volume-control/volume-control.js
```diff
@@ -0,0 +1,130 @@
...
+
+    /**
+     * Adjusts the volume by a percent value.
+     *
+     * @param {number} percentage - The value.
+     * @returns {void}
+     */
+    _adjustVolume(percentage) {
+        this._getVolume().then(value => {
+            const newVolume = Math.min(100, value * (100 + percentage) / 100);
```
virtuacoplenny: If value is ever zero then I think volume would never decrease,

spot-client/src/common/css/modal.scss
zbettenbuk: I agree, but I think in addition we should blur the background or apply some tinting. Otherwise the entire screen just looks like one big mess of buttons. What do you think?

E.g.:
image

zbettenbuk: Especially on a mobile screen:

image

virtuacoplenny: I remembered the UX design now. It’s intentional that the other buttons be clickable while the modal is displayed.

spot-electron/src/volume-control/volume-control.js
zbettenbuk: Right... I think I was drunk when I wrote this

virtuacoplenny:
I had made a PR to add the feature check for electron but Pawel prefers we do it in this PR: https://github.com/jitsi/jitsi-meet-spot/pull/309. Let me know how we should proceed.

spot-client/src/common/css/modal.scss
zbettenbuk: Well, ok then

spot-client/src/spot-remote/ui/components/remote-control-menu/menu-button.js
zbettenbuk: I tried, but that's just way too different. But let's see what Yana suggests about this. If she wants to go with a rounded button similar to the main nav buttons, we can replace them

spot-client/src/common/css/button.scss
zbettenbuk: Right! Also modified the nav container a bit, because it didn't scale nicely with the new button. I hope it's good now

spot-client/src/common/css/page-layouts/volume-modal.scss
zbettenbuk: Good point, I move them to the variables too. Thanks

spot-client/src/common/css/page-layouts/more-modal.scss
zbettenbuk: I agree, actually I had issues understanding how these modals work. Let's schedule that for later. It will be painful if we don't do it, so i'm sure we won't forget

spot-client/src/common/remote-control/remote-control-service.js
zbettenbuk: Probably. This is a matter of world view :) +1 and -1 makes the code easier inside the volume button and the signalling, but calling them 'up' and 'down' makes it a bit more abstract and makes the external API more human friendly. Both are good approaches, I think. Let's revisit this later, and change it if need be

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
zbettenbuk: Right, thanks!

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
zbettenbuk: Added there, thanks!

spot-electron/src/volume-control/volume-control.js
virtuacoplenny: I meant it would never increase in volume, not never decrease. I must have had sympathy drunkenness.

spot-electron/src/application-menu/applicationmenu.js
zbettenbuk: That was a remainder of a debugging.... Remover, thanks

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
zbettenbuk: Yes, we can do that. Although I'd be worried about promises that may never be resolved, for example when we try to send a message to a non-electron client. Although, since it's not really an async flow, I'd go with a message-reply approach with listener for now, if need be

spot-client/src/spot-remote/ui/components/remote-control-menu/menu-button.js
virtuacoplenny: Alright, I'll give a shot at refactoring the CSS a bit to support this. My mess, my burden.

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
zbettenbuk: This is something we should do when there is a need for that, I think

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
zbettenbuk: Ticket filed

spot-client/src/spot-tv/native-functions/native-controller/native-controller.js
zbettenbuk: Ticket filed

spot-client/src/spot-remote/ui/components/remote-control-menu/volume-button.js
zbettenbuk: Yeah, since we're probably going to go for the slider thing, I wouldn't improve this screen too much now

spot-client/src/spot-remote/ui/components/remote-control-menu/volume-button.js
zbettenbuk: Ok, I rewrite that

PR #298, opened 2019-05-16
Bind webpack's dev server to 0.0.0.0

paweldomas:
Binding on all interfaces makes it possible to display the page on LAN

### DIFF
--- a/spot-client/webpack.config.js
+++ b/spot-client/webpack.config.js
@@ -16,6 +16,7 @@ module.exports = {
         compress: true,
         contentBase: path.join(__dirname, '/'),
         historyApiFallback: true,
+        host: '0.0.0.0',
         port: 8000,
         publicPath: '/dist/'
     },


spot-client/webpack.config.js
```diff
@@ -16,6 +16,7 @@ module.exports = {
         compress: true,
         contentBase: path.join(__dirname, '/'),
         historyApiFallback: true,
+        host: mode === 'development' ? '0.0.0.0' : 'localhost',
```
virtuacoplenny: Is the ternary necessary? I'm wondering when the mode wouldn't be development. Do the webdriver tests still pass? They might be defaulted to localhost.

paweldomas: Yeah probably it's not, at first it was not there, but then I have spotted the mode and that made me think it can be only in dev.

I will check on the tests

paweldomas: but that devServer is used only in dev mode anyway ? is that it ?

paweldomas: `npm run test` passes

virtuacoplenny: It's used only for launching the dev server, correct. The mode is used to for webpack to do extra work, like uglifying when it is set to production.

virtuacoplenny: > `npm run test` passes

I meant the tests in spot-webdriver/, not the unit tests in spot-client/.

virtuacoplenny: K, I checked, looks like. I clearly don't understand how this "host" field works.

paweldomas: No it's not working for localhost now...

paweldomas: I am about to close it. I don't know if this worked with the ternary, but will try now if not will close the PR

virtuacoplenny: Are you sure you want to close? I think it's fine as long as we still have a convenient way to run the server and tests locally.

paweldomas: I would have to find a way to direct the test to 0.0.0.0 instead of localhost

virtuacoplenny: The tests default to localhost; they don't require it. Look for this line:
`const TEST_SERVER_URL = process.env.TEST_SERVER_URL || 'http://localhost:8000';`

paweldomas: oh thanks, I couldn't find that

paweldomas: hmm I think I'm just dump, I didn't realized that webpack must be started manually for the tests

paweldomas: yes, the tests are working as is

virtuacoplenny: That is correct. We do not have a script that starts the server and runs the tests--something we'll need to re-enable jenkins testing. Right now the tests take in whatever URL and run against that.

paweldomas: yes, that makes sense

PR #299, opened 2019-05-16
feat: add a help icon to the join code entry page

paweldomas:
Clicking on the help icon takes user to the help page.

Image from iOS (1)
Image from iOS
Screenshot 2019-05-16 14 52 41
Screenshot 2019-05-16 14 54 02

### DIFF
--- a/spot-client/src/spot-remote/ui/views/help.js
+++ b/spot-client/src/spot-remote/ui/views/help.js
@@ -1,0 +1,88 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+import { ROUTES } from 'common/routing';
+import { Button, View } from 'common/ui';
+import { windowHandler } from 'common/utils';
+
+/**
+ * Displays a view with basic usage instructions for Spot Remote.
+ *
+ * @extends React.Component
+ */
+class Help extends React.Component {
+    static propTypes = {
+        history: PropTypes.object
+    };
+
+    /**
+     * Gets the URL which point to the join code entry page.
+     *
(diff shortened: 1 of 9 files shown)


spot-client/src/spot-remote/ui/views/help.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    /**
+     * Takes the user to the join code entry page.
+     *
+     * @private
+     * @returns {void}
+     */
+    _onOkButtonClicked() {
+        this.props.history.push(ROUTES.CODE);
```
virtuacoplenny: I didn't know Route would automatically pass in history. You don't want to use the history import instead?

paweldomas: Hmm I saw it's used through the props in other components, so I wanted to be consistent.

virtuacoplenny: I created the history import somewhat recently so it can be used in actions/middleware. I broke consistency and never converted all the other instances. Do you prefer using it through props? Or is it just consistency that matters to you?

paweldomas: My understanding was to use the history import from actions/middleware, but try to use history property from components ?

virtuacoplenny: It should just work. It's the same object that allows for talking indirectly to window.history.

paweldomas: yes

PR #300, opened 2019-05-16
Device selection preview fixes

virtuacoplenny:


spot-client/src/spot-tv/ui/components/setup/select-media/select-media.js
```diff
@@ -244,8 +244,9 @@ class SelectMedia extends React.Component {
     _onDeviceListChange(devices) {
         const newDeviceLists = this._getDefaultDeviceListState();
 
-        devices.forEach(({ kind, label }) => {
+        devices.forEach(({ deviceId, kind, label }) => {
             const formatted = {
+                deviceId,
```
zbettenbuk: What are you doing with this field?

virtuacoplenny: So that is the field in the object that gets passed into the preview components. The preview components look in the object, for which I really need to make a jsdoc type, and grab the device ID and call getUserMedia with it. Without it the previews will be calling getUserMedia and allowing the browser to choose the device.

Removal of deviceId occurred when I formatted the object to custom (versus what's returned from enumerateDevice) so I could supported a "None" option. So proper device previews has been broken since then.

There are some other edge cases around previews I'd like to tighten up too, like no deviceId showing no preview and maybe handling disconnects.

PR #303, opened 2019-05-17
ref(help): remove link

paweldomas:
The link can be clicked from the iPad app and the user will end up on the unsupported page with no return. Remove the link to remove the problem.

### DIFF
--- a/spot-client/src/spot-remote/ui/views/help.js
+++ b/spot-client/src/spot-remote/ui/views/help.js
@@ -42,8 +42,6 @@ class Help extends React.Component {
      * @inheritdoc
      */
     render() {
-        const spotTvHomeUrl = Help._getSpotTvUrl();
-
         return (
             <View name = 'help'>
                 <div className = 'help-view'>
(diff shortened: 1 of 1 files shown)


spot-client/src/spot-remote/ui/views/help.js
```diff
@@ -54,13 +52,8 @@ class Help extends React.Component {
...
                             This app is a remote controller for a conference room.
                             <div>
-                                Open&nbsp;
-                                <a
-                                    className = 'spot-home-link'
-                                    href = { spotTvHomeUrl } >
-                                    { spotTvHomeUrl }
-                                </a>
-                                &nbsp;in a browser to setup your conference room and obtain a share key.
+                                Open&nbsp;{ Help._getSpotTvUrl() }&nbsp;in a browser to setup your conference room
```
virtuacoplenny: The non-breaking spaces might be removable.

paweldomas: yes

PR #306, opened 2019-05-18
ref(css): use sass partials to avoid duplicate imports

virtuacoplenny:
sass will import a file for each @import declaration,
and multiple @import declarations for the same file
results in multiple copies of the file being bundled.
The workaround is make everything a partial that gets
blasted into a single file, in this case a new
index.scss file.

This current saves about .02mb (20kb) off the bundle.


paweldomas:
maybe it's a bit less readable with no explicit imports in each file, but I'm not doing much CSS here

virtuacoplenny:
Force pushing to resolve conflicts with more_modal.scss in master.

PR #309, opened 2019-05-20
feat(detection): add ability to check for volume control support

virtuacoplenny:
I did some mind-changing with naming mid-way through but maybe it works out better this way.

### DIFF
--- a/spot-client/src/common/app-state/spot-tv/selectors.js
+++ b/spot-client/src/common/app-state/spot-tv/selectors.js
@@ -57,3 +57,14 @@ export function getRemoteSpotTVRoomName(state) {
 export function isConnectedToSpot(state) {
     return Boolean(state.spotTv.spotId);
 }
+
+/**
+ * A selector which returns whether or not the Spot-TV has volume control
+ * support.
+ *
+ * @param {Object} state - The Redux state.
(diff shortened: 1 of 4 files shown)


spot-client/src/common/app-state/spot-tv/selectors.js
```diff
@@ -57,3 +57,14 @@ export function getRemoteSpotTVRoomName(state) {
...
 }
+
+/**
+ * A selector which returns whether or not the Spot-TV has volume control
+ * support.
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {boolean}
+ */
+export function isVolumeControlSupported(state) {
```
paweldomas: is that used anywhere ?

virtuacoplenny: Not right now. It's for the volume control work that should be landing this week.

paweldomas: Can we land it with that work as first commit in the PR ?

virtuacoplenny: That’s fair. I’ll leave it I zoli.

PR #314, opened 2019-05-22
 feat(detection): add ability to check for volume control support

virtuacoplenny:
Re-opening of #309. These changes didn't make it in before volume control so they're being applied after volume control has been merged. The volume control button simply won't display.

We'll need to figure out the UI/UX for this case a big better. But at least for now we're not gonna show the volume buttons when they won't do anything.
Screen Shot 2019-05-22 at 12 43 15 PM

### DIFF
--- a/spot-client/src/index.js
+++ b/spot-client/src/index.js
@@ -40,7 +40,7 @@ const store = createStore(
         spotTv: {
 
             // Will get overridden on Spot-Remote by Spot-TV updates.
-            supportsOSFunctionality: isElectron()
+            electron: isElectron()
         },
         ...getPersistedState()
     },
--- a/spot-client/src/index.js
+++ b/spot-client/src/index.js
@@ -24,6 +24,7 @@ import {
 import {
     getDeviceId,
     getPersistedState,
+    isElectron,
     setPersistedState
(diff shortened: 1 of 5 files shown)


spot-client/src/index.js
```diff
@@ -36,6 +37,11 @@ const store = createStore(
         config: {
             ...setDefaultValues(window.JitsiMeetSpotConfig)
         },
+        spotTv: {
+
+            // Will get overridden on Spot-Remote by Spot-TV updates.
+            electron: isElectron()
```
paweldomas: should that be defined for the spot remote at all ?

paweldomas:
It says there are conflicts...

virtuacoplenny:
Fixed conflicts from restyling the volume button in more-modal.js.

yanas:
If we detect that only 1 item is available in More shouldn't we display that Item at the place of More?

virtuacoplenny:
> If we detect that only 1 item is available in More shouldn't we display that Item at the place of More?

If it's known we can encounter that case of having only one item in the More Menu then yes I agree. Right now I don't know what the plan is so I'd be hesitant to add that logic.

PR #319, opened 2019-05-24
ref(controller): rename folder to spot-controller

virtuacoplenny:


virtuacoplenny:
Force pushed to fix conflicts with the root README.

README.md
```diff
@@ -4,6 +4,7 @@ This is the mono-repo for jitsi meet's room integration, named Spot.
 
 in-room-controller/ contains the Android and iOS clients to be used as a permanent in-room remote
 spot-admin/ contains a mock server for testing integration between Spot and a backend service
+spot-controller/ contains the Android and iOS clients to be used as a permanent in-room remote
```
zbettenbuk: Ami I right assuming that this will also be the standalone store-version of the remote, so then not only the in-room controller?

virtuacoplenny: I didn't know there was some other intent of the react-native app. I did a bad conflict fix for this so I have to change it anyway. I'll remove the in-room part.

paweldomas: should the description at line 5 for `in-room-controller/` directory be removed  ?

virtuacoplenny: Yes, that was the bad conflict fix I was talking about.

PR #324, opened 2019-05-26
fix(analytics): add event for toggling tile view

virtuacoplenny:


virtuacoplenny:
Force pushed because I accidentally added another commit to this branch.

PR #325, opened 2019-05-26
feat(tests): add test for checking no volume control on browsers

virtuacoplenny:
Not sure how to test for the existence of volume controls,
which is gated for electron only.

Edit: looked around a bit and there's electron-chromedriver for testing electron apps. Probably that'll need to be done one day to test volume control.


virtuacoplenny:
Force pushed tor resolve conflicts with adding the new qa-id to the volume button.

virtuacoplenny:
Force pushing because github said there's a conflict but git handled it without intervention.

PR #326, opened 2019-05-26
Move some calls to RCS into actions

virtuacoplenny:


spot-client/src/common/app-state/remote-control-service/actions.js
```diff
@@ -209,6 +209,21 @@ export function startWirelessScreensharing() {
...
+ * Cleans ups any wireless screensharing connection being established or already
+ * established.
+ *
+ * @returns {Function}
+ */
+export function forceStopWirelessScreenshare() {
+    return dispatch => {
+        remoteControlService.destroyWirelessScreenshareConnections();
+
+        dispatch(setLocalWirelessScreensharing(false));
```
paweldomas: does it duplicate the code from `stopScreenshare` ?

virtuacoplenny: stopScreenshare calls destroyWireless. stopScreenshare does the XMPP command messaging, whereas destroyWireless does not so its cleanup is all local. I don't know why I thought the difference was important, maybe it isn't.

paweldomas: I mean those two lines 222 and 223

virtuacoplenny: Same answer. I'll have to look into it more.

virtuacoplenny:
Force pushed to resolve conflicts.

spot-client/src/common/app-state/remote-control-service/actions.js
virtuacoplenny: Answer is no they do different things but it can probably be refactored so there isn't redundancy.

paweldomas: yes I know they do different things but the logic for resetting the state is duplicated

virtuacoplenny: It looks like local wireless screenshare state is in one reducer while spot-tv state is another reducer. setLocalWirelessScreensharing updates local state while setSpotTVState is called to fake an immediate response from Spot-TV about its state update. I'm going to remove the setSpotTVState call because it should not be needed for this case.

PR #327, opened 2019-05-27
ref(rcs): rename subclasses to RCServer and RCClient

virtuacoplenny:
This is phase two of a refactor effort to separate spot-remote and spot-tv remote control service logic. Phase 1 was #318, splitting remote control service into two separate objects. Phase 3 will be moving files around.

This PR is built on top of #326, which has some moving of remote control service actions.

### DIFF
--- a/spot-client/src/common/remote-control/constants.js
+++ b/spot-client/src/common/remote-control/constants.js
@@ -1,6 +1,6 @@
 /**
- * An enumeration of supported commands Spot can processes. All values are from
- * events triggered by the jitsi meet iframe api.
+ * An enumeration of supported commands {@code RemoteControlServer} can process.
+ * All values are from events triggered by the Jitsi-Meet iFrame API.
  */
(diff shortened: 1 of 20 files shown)


spot-client/src/common/remote-control/constants.js
```diff
@@ -108,14 +117,14 @@ export const SERVICE_UPDATES = {
...
+     * A command from a {@code RemoteControlClient} has been received.
      */
-    SPOT_REMOTE_MESSAGE_RECEIVED: 'message-received',
+    CLIENT_MESSAGE_RECEIVED: 'message-received',
 
     /**
      * The remote control service has received an updated Spot-TV state.
      */
-    SPOT_TV_STATE_CHANGE: 'spot-tv-state-change',
+    SERVER_STATE_CHANGE: 'remote-control-server--state-change',
```
paweldomas: double `-` in the value ?

PR #329, opened 2019-05-29
Soft beacon for electron

zbettenbuk:
This is the 1st iteration for the soft beacon support for the electron app. This feature starts a soft beacon and starts to emit the join code using the major and minor version of the beacon. The region is fixed, so then we can listen to that explicit region in the mobile app even when the app is not running.

NOTE:
1. The solution only supports mac (mojave) for now, more platforms to coma later.
2. For testing, to resolve the join code from the versions one needs to use this code:
```parseInt((majorInt).toString(16) + (minorInt).toString(16), 16).toString(36)```
This logic will be embedded into the mobile app in iteration 2

### DIFF
--- a/spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
+++ b/spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
@@ -1,0 +1,18 @@
+import { SPOT_TV_SET_JOIN_CODE } from 'common/app-state';
+import { MiddlewareRegistry } from 'common/redux';
+
+import { nativeController } from '../native-controller';
+
+MiddlewareRegistry.register(() => next => action => {
+    const result = next(action);
+
+    switch (action.type) {
+    case SPOT_TV_SET_JOIN_CODE:
+        nativeController.sendMessage('updateJoinCode', {
+            joinCode: action.joinCode
+        });
+        break;
+    }
+
+    return result;
+});
--- a/spot-electron/config.js
+++ b/spot-electron/config.js
@@ -3,6 +3,17 @@ const process = require('process');
 module.exports = {
 
     /**
+     * Config values for the beacon functionality.
+     */
+    beacon: {
+
+        /**
+         * Region ID for the beacon. This is permanent and must be unique for custom deployments.
+         */
+        region: 'bf23c311-24ae-414b-b153-cf097836947f'
+    },
+
+    /**
      * The default URL to connect to.
      */
     defaultSpotURL: process.env.SPOT_URL || 'https://spot.jitsi.net/spot',
--- a/spot-electron/package.json
+++ b/spot-electron/package.json
@@ -12,6 +12,7 @@
         "babel-eslint": "10.0.1",
         "electron": "5.0.1",
         "electron-builder": "20.40.2",
+        "electron-rebuild": "1.8.5",
         "eslint": "5.16.0",
         "eslint-config-jitsi": "github:jitsi/eslint-config-jitsi#1.0.1",
         "eslint-plugin-flowtype": "3.8.2",
@@ -26,15 +27,17 @@
         "dist": "electron-builder",
         "lint": "eslint .",
         "pack": "electron-builder --dir",
+        "postinstall": "electron-rebuild",
         "start": "electron ."
     },
     "version": "1.0.0",
     "dependencies": {
         "jitsi-meet-logger": "github:jitsi/jitsi-meet-logger#5da3033d647054457597920252b3387fa43a17fa",
         "lodash": "4.17.11",
-        "node-osascript": "^2.1.0"
+        "node-osascript": "2.1.0"
     },
     "optionalDependencies": {
-        "win-audio": "2.0.2"
+        "win-audio": "2.0.2",
+        "xpc-connect": "git+https://git@github.com/jitsi/xpc-connect.git#33fd79673c57f15167067778fc6ade0a4f46da07"
     }
 }
--- a/spot-electron/src/bt-beacon/btbeacon.js
+++ b/spot-electron/src/bt-beacon/btbeacon.js
@@ -1,0 +1,61 @@
+const { config } = require('../config');
+
+const { clientController } = require('../client-control');
+const { logger } = require('../logger');
+const { joinCodeToVersion } = require('../utils');
+
+const daemon = require('./daemon');
+
+/**
+ * Represents a soft-beacon emulated by the computer running Spot TV
+ */
+class BTBeacon {
+
+    /**
+     * Instantiates a new instance of the class.
+     */
+    constructor() {
+        this.region = config.getValue('beacon.region');
+
+        if (daemon && this.region) {
+            daemon.connect().then(() => {
+                clientController.on('updateJoinCode', ({ joinCode }) => {
+                    this.joinCodeUpdated(joinCode);
+                });
+            });
+        } else {
+            logger.warn('BT beacon functionality is not available.');
+        }
+    }
+
+    /**
+     * Updates the beacon to emit the newly received join code.
+     *
+     * @param {string} joinCode - The join code to use as major and minor version for the beacon.
+     * @returns {void}
+     */
+    joinCodeUpdated(joinCode) {
+        if (this.joinCode === joinCode) {
+            return;
+        }
+
+        logger.info('Updating BT beacon...');
+        if (joinCode) {
+            const majorMinorVersion = joinCodeToVersion(joinCode);
+
+            daemon.stopBeacon().then(() => {
+                daemon.startBeacon(this.region, majorMinorVersion[0], majorMinorVersion[1]).then(() => {
+                    this.joinCode = joinCode;
+                    logger.info(`BT Beacon updated with join code ${joinCode}`);
+                });
+            });
+        } else {
+            // No join code, we disable the beacon.
+            daemon.stopBeacon().then(() => {
+                logger.info('BT beacon stoppped.');
+            });
+        }
+    }
+}
+
+module.exports = new BTBeacon();
--- a/spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
+++ b/spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
@@ -1,0 +1,159 @@
+const XpcConnection = require('xpc-connect');
+
+const { logger } = require('../../../logger');
+
+const DEVICE_STATES = [ 'Unknown', 'Resetting', 'Unsupported', 'Unauthorized', 'Powered Off', 'Powered On' ];
+
+/**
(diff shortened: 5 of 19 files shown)

### CODE REVIEW

File: spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js [hunk 1]
```diff
@@ -0,0 +1,27 @@
+import { SPOT_TV_SET_JOIN_CODE } from 'common/app-state/spot-tv/action-types';
```
virtuacoplenny: Why did you choose to import from the action-types directory instead of getting from common/app-state?

spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
```diff
@@ -0,0 +1,27 @@
...
+    return result;
+});
+
+/**
+ * Updates the native wrapper about a join code change.
+ *
+ * @param {string} joinCode - The join code that the spot TV instance is using now.
+ * @returns {void}
+ */
+function _updateJoinCode(joinCode) {
```
virtuacoplenny: Style preference comment: I wonder if you would prefer inlining this.

spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
```diff
@@ -0,0 +1,27 @@
...
+        _updateJoinCode(action.joinCode);
+    }
+
+    return result;
+});
+
+/**
+ * Updates the native wrapper about a join code change.
+ *
+ * @param {string} joinCode - The join code that the spot TV instance is using now.
```
virtuacoplenny: "Spot-TV" is generally how I've been spelling it when I remember I should be consistent.

spot-electron/config.js
```diff
@@ -2,6 +2,17 @@ const process = require('process');
...
 
+    /**
+     * Config values for the beacon functionality.
+     */
+    beacon: {
+
+        /**
+         * Region ID for the beacon. This is permanent and must be unique for custom deployments.
+         */
+        region: 'bf23c311-24ae-414b-b153-cf097836947f'
```
virtuacoplenny: Do you know what the plan might be for deployments to use? A backend will provide the beacon and clients will do some exchange with the backend?

spot-electron/package.json
```diff
@@ -26,15 +27,17 @@
...
     "dependencies": {
         "jitsi-meet-logger": "github:jitsi/jitsi-meet-logger#5da3033d647054457597920252b3387fa43a17fa",
         "lodash": "4.17.11",
-        "node-osascript": "^2.1.0"
+        "node-osascript": "2.1.0"
     },
     "optionalDependencies": {
-        "win-audio": "2.0.2"
+        "win-audio": "2.0.2",
+        "xpc-connect": "git+https://git@github.com/jitsi/xpc-connect.git#33fd79673c57f15167067778fc6ade0a4f46da07"
```
virtuacoplenny: Should we wait a little bit to see what the response is from the PR?

spot-electron/src/bt-beacon/btbeacon.js
```diff
@@ -0,0 +1,61 @@
...
+        this.region = config.getValue('beacon.region');
+
+        if (daemon && this.region) {
+            daemon.connect().then(() => {
+                clientController.on('updateJoinCode', ({ joinCode }) => {
+                    this.joinCodeUpdated(joinCode);
+                });
+            });
+        } else {
+            logger.warn('BT beacon functionality is not available.');
```
virtuacoplenny: From my knowledge, which is years old, windows machines might not have bluetooth.  Should there be something exposed in the Spot-TV UI indicating lack of bluetooth support or that bluetooth is being broadcast? I can add it if need be but I don't know if there is a plan already to add it.

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
```diff
@@ -0,0 +1,159 @@
...
+        return new Promise(resolve => {
+            this.readyPromise = resolve;
+            this.xpcConnection.setup();
+            this.sendMessage(1, {
+                kCBMsgArgName: `jitsi-bt-beacon-${(new Date()).getTime()}`,
+                kCBMsgArgOptions: {
+                    kCBInitOptionShowPowerAlert: 1
+                },
+                kCBMsgArgType: 1,
+                kCBMsgArgVersion: 20161219
```
virtuacoplenny: There are a lot of magical numbers in this file. Would someone working on bluetooth be able to quickly get context on what these numbers mean?

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Connects to the BT device (opens XPC channel).
+     *
+     * @returns {Promise<void>}
+     */
+    connect() {
+        return new Promise(resolve => {
+            this.readyPromise = resolve;
+            this.xpcConnection.setup();
+            this.sendMessage(1, {
```
virtuacoplenny: Why is the message id hardcoded as 1?

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
```diff
@@ -0,0 +1,159 @@
...
+    /**
+     * Updates the advertising status of the device.
+     *
+     * @param {boolean} status - The status to set.
+     * @returns {void}
+     */
+    setAdvertisingStatus(status) {
+        this.advertisingStatus = status;
+        logger.info('BluetoothD advertising', this.advertisingStatus ? 'started' : 'stopped');
+        if (typeof this.advertisingStatusUpdatedPromise === 'function') {
```
virtuacoplenny: When would a promise be a typeof function?

spot-electron/package.json
zbettenbuk: Not sure, because I also changed the package.json, that I think they don't want to merge

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
zbettenbuk: Not really, these are API constants. I don't know what they mean and frankly don't even care :)

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
zbettenbuk: Message IDs are not sequence in case of XPC. Those actually denote the command we want XPC to execute. Message ID 1 is the command to power on the BT device

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
zbettenbuk: This is not a promise, this is the resolve function. Maybe needs a better naming?

spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
zbettenbuk: Dunno. I guess I thought it's logical :)

I'll fix this, thanks

spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
zbettenbuk: Sure!

spot-electron/config.js
zbettenbuk: Good question. This has to be fixed in order to allow the iOS app to detect it even when the app is not running. Yes, I think it should come from the deployment later. Is there already a way to do that or we need to wait for the backend?

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
virtuacoplenny: Yes, I would suggest renaming. Maybe denote it's a callback. Although I wonder then why have promises and callbacks.

spot-electron/src/bt-beacon/btbeacon.js
zbettenbuk: I don't exactly know. Yeah, I guess it would be great to have it. Let's add a story for that!

spot-electron/config.js
virtuacoplenny: There isn't a way I know of and I'm generally unfamiliar with the work/discussion done for bluetooth before you started working on it. (I saw the demo a while and that might be the sum of my knowledge.)

spot-electron/package.json
virtuacoplenny: Looks like the maintainer is okay with them. Let's wait a bit if that's okay. I'm assuming "a bit" is less than a couple days.

spot-electron/src/bt-beacon/btbeacon.js
virtuacoplenny: Okay, I'll make a ticket and just do it one day.

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
virtuacoplenny: Okay. Maybe when doing the windows work I'll have to look. I'll use constants for the numbers if I have to figure out what they mean.

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
virtuacoplenny: I wonder then if a rename is necessary to denote these are intentionally constant. But maybe converting the magic numbers to constants would resolve that.

spot-client/src/spot-tv/native-functions/bt-beacon/middleware.js
virtuacoplenny: Oh, I guess `SPOT_TV_SET_JOIN_CODE` doesn't get exported. I never had middleware planned for jitsi-meet-spot and never thought something outside of app-state would need the action type. So...undo? Or maybe think of how you would want it?

zbettenbuk: Nah, added the export. But I'm hesitating:
Exporting too much on the top level leads easily to circular deps. But it makes the code tidy. What do you think?

spot-electron/config.js
zbettenbuk: Ok, let's leave it here now, and we adapt when we finally have backend

spot-electron/package.json
zbettenbuk: Yeah, had to rebase now. We'll see

spot-electron/src/bt-beacon/daemon/darwin/bluetoothd.js
zbettenbuk: Renamed

PR #330, opened 2019-05-29
feat(meeting): log device errors reported by meeting frame

virtuacoplenny:
Relies on https://github.com/jitsi/jitsi-meet/pull/4289 being deployed. One day these device errors will be reported to a backend so admins know about any device errors the Spot-TV is encountering. For now they get dumped into logs only.

### DIFF
--- a/spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
+++ b/spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
@@ -473,7 +473,7 @@ export class MeetingFrame extends React.Component {
      * @returns {void}
      */
     _onReportDeviceError(type, message) {
-        logger.log('device error occurred within the meeting', {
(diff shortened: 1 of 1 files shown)


spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -460,6 +464,21 @@ export class MeetingFrame extends React.Component {
...
+    /**
+     * Callback invoked when Jitsi-Meet failed to use a media device.
+     *
+     * @param {string} type - The type of the error.
+     * @param {string} message - Additional information about the error.
+     * @private
+     * @returns {void}
+     */
+    _onReportDeviceError(type, message) {
+        logger.log('device error occurred within the meeting', {
```
zbettenbuk: Would it be good to use error here?

PR #335, opened 2019-06-01
feat(connection): show overlay during auto reconnect

virtuacoplenny:
Using the common animated three dots.

Screen Shot 2019-05-31 at 5 43 42 PM
Screen Shot 2019-05-31 at 5 59 35 PM


paweldomas:
Hmm I was thinking we would go back to the three dots and hide all the UI, possibly through the abstract loader and withRemoteControl wrapper. Is it that this behavior was no longer desired and the overlay is preferred ?

virtuacoplenny:
There is no set desired behavior right now. I thought this case should be unique and quick enough to recover from that causing an unmount of the view (for example the meeting frame) would not be desired.

PR #336, opened 2019-06-01
fix(calendar): search more places for the meeting url

virtuacoplenny:
Relies on https://github.com/jitsi/js-utils/pull/9.


spot-client/src/spot-tv/calendars/google.js
```diff
@@ -129,12 +134,21 @@ export default {
  * @param {Array<Object>} events - The calendar events to filter.
  * @param {string} calendarEmail - The email of the calendar configured to
  * display.
+ * @param {Array<string>} knownDomains - Whitelist of domains which can be used
+ * as meeting links.
  * @returns {Array<Object>}
  */
-function filterJoinableEvents(events = [], calendarEmail) {
+function filterJoinableEvents(events = [], calendarEmail, knownDomains) {
```
paweldomas: is this currently only to be used by google calendar ? would it be possible to have the detection in the calendar service instead and handle any calendar ?

But on the other hand probably the backend calendar will have a dedicated field, so not sure. Maybe as opt-in ? If it doesn't make sense for outlook or if you don't feel like doing it's fine too.

paweldomas: Oh I see it's used in outlook too.

spot-client/src/spot-tv/calendars/google.js
```diff
@@ -129,12 +134,21 @@ export default {
...
     return events.map(event => {
-        const { attendees, location, end, id, start, summary } = event;
-        const meetingUrl = getMeetingUrl(location);
+        const { attendees, end, id, start, summary } = event;
+        const meetingUrl = getMeetingUrl([
```
paweldomas: It feels like it's a bit of a duplication to do this same work in all of the subclasses. How about passing the fields to search to the service and do the job in one place ?

virtuacoplenny: The work happens in each implementation of a calendar because the calendar services return events in different formats, and it is in this function that the events get out into a standard format. When searching for a meeting url, differnt fields are being searched between google and outlook. Does that change your opinion? If not I might have misunderstand your suggestion.

paweldomas: Yeah I understand that they use different fields for search and my suggestion was to pass those fields that are to be search to the calendar service for processing.

virtuacoplenny: I implemented passing the fields.

PR #342, opened 2019-06-03
Move connect logic from component to actions and other refactoring

paweldomas:
@virtuacoplenny probably this could be a few separate PRs. Let me know if you want it this way.
I'm opening as a draft PR, because I have not tested the changes after rebasing and splitting them into separate commits.


spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -22,41 +70,167 @@ import {
...
+            getState
+        });
+
+        const state = getState();
         const { joinCodeServiceUrl } = getSpotServicesConfig(getState());
 
         return remoteControlClient.exchangeCode(joinCode, { joinCodeServiceUrl })
-            .then(roomInfo => {
+            .then(roomInfo => remoteControlClient.connect({
+                autoReconnect: true,
```
virtuacoplenny: Is autoReconnect ever false? It might have existed as a temporary variable while moving reconnect logic into RCS. Probably a future cleanup you or I should do.

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -1,10 +1,15 @@
 import {
```
virtuacoplenny: Side thought: I keep getting torn between if we should have a lot of this logic in actions or middleware. Like have middleware that essentially sets up pubsub to RCS but keep the logic in RCS; do not use middleware to fire additional actions from one action but rather do like connect() on an explicit CONNECT action type.. I think the two implementations would actually look very similar but the dependency relationships might look

spot-client/src/spot-remote/ui/loaders/withRemoteControl.js
```diff
@@ -124,117 +42,17 @@ export class RemoteControlLoader extends AbstractLoader {
...
-        roomInfo: getRoomInfo(state),
-        remoteControlConfiguration: getRemoteControlServerConfig(state)
-    };
-}
 
-const ConnectedRemoteControlLoader
-    = withRouter(connect(mapStateToProps)(RemoteControlLoader));
+const ConnectedRemoteControlLoader = withRouter(RemoteControlLoader);
 
 export default generateWrapper(ConnectedRemoteControlLoader);
```
virtuacoplenny: Is withRemoteControl still necessary to exist?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -72,12 +76,17 @@ export class BaseRemoteControlService extends EventEmitter {
             onPresenceReceived: this._onPresenceReceived
         });
 
-        this.xmppConnectionPromise = this.xmppConnection.joinMuc({
+        this.xmppConnectionPromise = this.exchangeCode(
```
virtuacoplenny: Nice. Good idea.

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -86,40 +86,33 @@ export function connectToSpotTV(joinCode, shareMode) {
...
-                });
-
-                // In the wrong password case return back to join code entry.
-                if (error === 'not-authorized') {
-                    dispatch(addNotification('error', 'Invalid share key'));
-                } else if (error === 'unrecoverable-error') {
-                    dispatch(addNotification('error', 'Something went wrong'));
-                }
-
-                _onDisconnected(error, {
```
virtuacoplenny: You don't need this call to _onDisconnected anymore?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -186,37 +186,54 @@ export class BaseRemoteControlService extends EventEmitter {
...
-            logger.log(`Will use ${joinCodeServiceUrl} to validate the join code...`);
-
-            return fetchRoomInfo(joinCodeServiceUrl, code)
-                .then(({ roomName, roomLock }) => {
-                    return {
-                        roomName,
-                        roomLock
-                    };
-                });
+            return this.exchangeCodeWithBackend(code.trim(), joinCodeServiceUrl);
```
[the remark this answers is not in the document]
paweldomas: I was about to add join code strategies as a next step(strategy design pattern).

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -186,37 +186,54 @@ export class BaseRemoteControlService extends EventEmitter {
...
+     *
+     * @param {string} code - The join code to exchange for connection information.
+     * @param {string} joinCodeServiceUrl - The URL pointing to the join code service.
+     * @returns {Promise<RoomInfo>} Resolve with join information or an error.
+     */
+    exchangeCodeWithBackend(code, joinCodeServiceUrl) {
+        logger.log(`Will use ${joinCodeServiceUrl} to validate the join code...`);
+
+        return fetchRoomInfo(joinCodeServiceUrl, code)
+            .then(({ roomName, roomLock }) => {
```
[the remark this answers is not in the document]
paweldomas: No, it's just that. Why ?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -186,37 +186,54 @@ export class BaseRemoteControlService extends EventEmitter {
...
-                roomLock: enteredCode.substring(3, 6)
+    /**
+     * Converts a join code into XMPP MUC credentials using a backend service.
+     *
+     * @param {string} code - The join code to exchange for connection information.
+     * @param {string} joinCodeServiceUrl - The URL pointing to the join code service.
+     * @returns {Promise<RoomInfo>} Resolve with join information or an error.
+     */
+    exchangeCodeWithBackend(code, joinCodeServiceUrl) {
+        logger.log(`Will use ${joinCodeServiceUrl} to validate the join code...`);
```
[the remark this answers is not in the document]
paweldomas: right

spot-client/src/spot-remote/app-state/actions.js
paweldomas: true

spot-client/src/spot-remote/app-state/actions.js
paweldomas: yeah, I guess we could do that

spot-client/src/spot-remote/ui/loaders/withRemoteControl.js
paweldomas: I was thinking this component could be useful to switch to the spinner while the connection is lost, but probably right now that's not the case, right ?

spot-client/src/spot-remote/app-state/actions.js
paweldomas: oh did it slip through in some of the commits ? No, it's not needed after your change to call disconnected when the connect promise is rejected. Will double check.

spot-client/src/common/remote-control/BaseRemoteControlService.js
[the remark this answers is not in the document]
virtuacoplenny: I wondered why you were destructring the result (`{ roomName, roomLock }`) instead of returning it directly. I thought maybe something else was being returned in the object.

spot-client/src/spot-remote/ui/loaders/withRemoteControl.js
virtuacoplenny: Okay, let's keep it around until we finalize between us what the desired behavior should be there. I suppose it's still useful to keep around to show the loading state while an initial connection is being made.

spot-client/src/spot-remote/app-state/actions.js
paweldomas: Okay, so it changed when the exchangeCode call was moved to be part of the connect. Because when it's part of connect the RCS will fire disconnected when it fails. Prior to that it needed to be called no `exchangeCode` rejection.

spot-client/src/common/remote-control/BaseRemoteControlService.js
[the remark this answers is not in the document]
paweldomas: Yeah, when I checked the code again I also started wondering. Fixed now

spot-client/src/spot-remote/app-state/actions.js
paweldomas: added as a separate commit

paweldomas:
Force pushed to squash the commits

paweldomas:
@virtuacoplenny force pushed to fix the conflicts

PR #348, opened 2019-06-07
Remote join code

paweldomas:
With the backend, Spot Remote will use different join code that Spot TV. After this commit a `joinCode` is the one used by the local instance to connect to the service and establish pairing between Spot TV and Spot Remote.

A remote pairing code is produced only by a Spot TV and displayed on the screen, so that a user can use it to pair a Spot Remote.


spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -105,8 +105,7 @@ export class BaseRemoteControlService extends Emitter {
     }
 
     /**
-     * Returns the current join code that is necessary to establish a connection
-     * to a Spot-TV.
+     * Returns the current join code that is necessary to establish a connection with the service.
```
virtuacoplenny: Let me know when you I can go through the docs (including the readme files) and I'll update everything as needed. Sometimes it's nice product something without having to think really hard.

PR #351, opened 2019-06-09
fix(reconnect): move not-authorized retry to XmppConnection

virtuacoplenny:


spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -151,7 +153,7 @@ export class BaseRemoteControlService extends Emitter {
         // wait a little bit to retry to avoid a stampeding herd
         const jitter = getJitterDelay();
 
-        const previousJoinCode = this.getJoinCode();
+        this._previousJoinCode = this.getJoinCode() || this._previousJoinCode;
```
paweldomas: why is this change necessary ?

virtuacoplenny: getJoinCode retrieves the muc + lock from the xmpp connection. If the connection does not exist an empty string is returned. On the first reconnect attempt this is fine. On the second reconect attempt, the xmpp connection has already been destroyed so getJoinCode will return an empty string. Would you prefer caching of the last join code instead of getJoinCode retrieving from the xmpp connection each time?

virtuacoplenny: Related question: should disconnect() ever be called during the reconnect flow?

paweldomas: >getJoinCode retrieves the muc + lock from the xmpp connection. If the connection does not exist an empty string is returned. On the first reconnect attempt this is fine. On the second reconect attempt, the xmpp connection has already been destroyed so getJoinCode will return an empty string. Would you prefer caching of the last join code instead of getJoinCode retrieving from the xmpp connection each time?

I think I have this change or was planning to do as part of the backend PR, so it is good for now.

> Related question: should disconnect() ever be called during the reconnect flow?

If it would emit disconnected events the it shouldn't. I'd say whatever is needed by the XMPP connection to re-establish the connection.

PR #355, opened 2019-06-11
ref(route): redirect to setup if backend is enabled

virtuacoplenny:


spot-client/src/spot-tv/ui/loaders/SpotTVRemoteControlLoader.js
```diff
@@ -47,15 +44,9 @@ export class SpotTVRemoteControlLoader extends React.Component {
...
-                // Connection in the backend mode is started by entering the pairing code on the setup screen
-                history.push(ROUTES.SETUP);
-            } else {
-                // In the no backend mode the connection logic loops forever, because as long as there are
-                // no network/config problems the connection must succeed.
-                this.props.dispatch(createSpotTVRemoteControlConnection({ retry: true }));
-            }
+            // In the no backend mode the connection logic loops forever, because as long as there are
+            // no network/config problems the connection must succeed.
+            this.props.dispatch(createSpotTVRemoteControlConnection({ retry: true }));
```
paweldomas: So this code should not be executed `if (this.props.isBackendEnabled)` or at least that was my original intention there. That top level retry loop should not kick in with backend, because the pairing code might be wrong for example and user needs to be taken back to setup.

For backend the connection should be created during the setup at least until there will be logic for restoring the backend registration and resuming the connection. It will probably be somewhere around here.

PR #362, opened 2019-06-12
fix(setup): store room name from backend

virtuacoplenny:


spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -48,8 +48,9 @@ export class SpotBackendService {
      */
     getRoomInfo() {
         return fetchRoomInfo(this.roomKeeperServiceUrl, this.getJwt())
-            .then(({ mucUrl }) => {
+            .then(({ mucUrl, name }) => {
                 return {
+                    name,
```
paweldomas: can you update the RoomInfo type's doc ? (in backend/utils)

paweldomas: no, wait it's somewhere else

paweldomas: in `BaseRemoteControlService`

paweldomas:
Also would be good to define a type for the room profile, but only if you feel like doing it

PR #365, opened 2019-06-13
feat(webview): cover in loading screen

virtuacoplenny:
third iteration:
no-launch

second iteration (outdated):
black-black

first iteration (outdated):
loading

### DIFF
--- a/spot-controller/App.js
+++ b/spot-controller/App.js
@@ -2,7 +2,6 @@ import AsyncStorage from '@react-native-community/async-storage';
 import React from 'react';
 import SideMenu from 'react-native-side-menu';
 
-import LoadingScreen from './src/LoadingScreen';
 import RemoteControl from './src/remote-control';
 import SettingsMenu from './src/settings-menu';
 import Setup from './src/setup';
@@ -27,7 +26,6 @@ export default class App extends React.Component {
         super(props);
 
         this.state = {
-            loading: __DEV__,
             remoteControlUrl: __DEV__ ? null : DEFAULT_URL
         };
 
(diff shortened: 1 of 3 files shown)


paweldomas:
Do you know what is the white screen saying Jitsi Meet Spot ?

spot-controller/App.js
```diff
@@ -26,6 +27,7 @@ export default class App extends React.Component {
         super(props);
 
         this.state = {
+            loading: __DEV__,
```
paweldomas: what does it do ? why set to `__DEV__` ?

virtuacoplenny: Right now only dev does something async (look in local storage). Whereas if it's not dev then it skips looking in local storage. So on dev what happens is the code entry screen displays, find a url or does not, and then displays the webview. But while it's reading from local storage, it shows the code entry screen.

In a follow up commit where I have to store if the app has ever been launched this.state.loading gets started as true, not __DEV__, and will be set to false after async storage has been searched for some flag saying the app has been launched before.

paweldomas: My understanding of the loading screen purpose was to show it during the time when the WebView is loading the page

virtuacoplenny: So you want me to remove hiding the flash of the code entry screen. I don't mind.

virtuacoplenny: I don't know if that's what you want but I went ahead and did it.

virtuacoplenny:
> Do you know what is the white screen saying Jitsi Meet Spot ?

The configured splash screen. I guess I forgot to ask: should the loading screen be white to match the splash screen?

virtuacoplenny:
Looks like launch screen is already removed in latest master. I've rebased.

PR #366, opened 2019-06-13
feat(join-code): redirect to help if query param present

virtuacoplenny:
But only do it one time. This becomes a makeshift
onboarding.


virtuacoplenny:
I should write some selenium tests for this.

PR #368, opened 2019-06-14
feat: add beacon client support

zbettenbuk:


spot-client/src/common/app-state/api/middleware.js
```diff
@@ -0,0 +1,48 @@
...
+    let target;
+
+    // React native webview
+    // eslint-disable-next-line prefer-const
+    target = target || window.ReactNativeWebView;
+
+    // Other (e.g. iFrame) implementations come here later
+    // ...
+
+    const apiMessage = JSON.stringify({
```
virtuacoplenny: Are you sure you need to stringify? On browsers-to-browser you don't need to.

spot-client/src/common/app-state/api/middleware.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {string} messageType - The message type.
+ * @param {string|Object} messageData - The message.
+ * @returns {void}
+ */
+function _sendApiMessage(messageType, messageData) {
+    let target;
+
+    // React native webview
+    // eslint-disable-next-line prefer-const
+    target = target || window.ReactNativeWebView;
```
virtuacoplenny: Is there a point to checking the truthiness of target?

spot-client/src/common/app-state/api/middleware.js
zbettenbuk: I couldn't get it done otherwise, but maybe I overlooked something

spot-client/src/common/app-state/api/middleware.js
zbettenbuk: I wanted to be consistent with the (not yet existing) other checks, so then they can be in any order. Do you think I shouldn't have it here yet?

zbettenbuk:
Discussed with Pawel, he trusts Lenny's review. He'll take a post-mortem look though

PR #370, opened 2019-06-16
feat(screenshare): new select designs

virtuacoplenny:
I didn't focus much on the narrow width (mobile) stylings. I just made sure they appear. One of the bigger deviations from the designs is not having a separate "modal" for when on an unsupported browser, as calling out "browser" doesn't work out well for the native apps. So instead for unsupported browsers show the wireless screenshare instructions instead of calling out an unsupported nature; this was a bug fix agreed upon several weeks ago with design anyway but I think it was forgotten about during the current screenshare design revisioning.

Screen Shot 2019-06-16 at 10 37 01 AM
Screen Shot 2019-06-16 at 10 36 56 AM
Screen Shot 2019-06-16 at 10 36 31 AM
Screen Shot 2019-06-16 at 10 37 04 AM
Screen Shot 2019-06-16 at 10 36 38 AM

Mobile ugliness. Narrow width needs work in general. Keep me honest and request any changes you think are necessary now, but we don't have any designs or design attention for most of narrow width.
Screen Shot 2019-06-16 at 10 36 04 AM

### DIFF
--- a/spot-client/src/common/css/_modal.scss
+++ b/spot-client/src/common/css/_modal.scss
@@ -22,17 +22,13 @@
         @media #{$mq-tablet} {
             height: auto;
             margin: 50px;
-            min-width: 500px;
             width: unset;
-        }
-
-        @media #{$mq-laptop-l} {
-            min-width: 550px;
         }
 
         .close {
             @include button-reset;
 
+            background-color: transparent;
             border: 0;
             position: absolute;
             top: 10px;
--- a/spot-client/src/common/css/_modal.scss
+++ b/spot-client/src/common/css/_modal.scss
@@ -42,7 +42,7 @@
 }
 
 .modal-content {
-    background-color: var(--container-bg-color);
+    background-color: var(--modal-bg-color);
     border-radius: 5px;
     display: flex;
     justify-content: center;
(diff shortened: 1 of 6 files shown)


yanas:
@virtuacoplenny a few questions:

- is the background opacity set to 80% instead of 70% ?
- the sharing url should be in bold
- would it be difficult to make this cover the whole screen (with opacity 1) on mobile and treat it as a separate screen, not a modal

virtuacoplenny:
> @virtuacoplenny a few questions:
> 
> * is the background opacity set to 80% instead of 70% ?
> * the sharing url should be in bold
> * would it be difficult to make this cover the whole screen (with opacity 1) on mobile and treat it as a separate screen, not a modal

Background opacity is currently 80. Am I supposed to take your CSS from zeplin and copy/paste it into the spot? I guess with each change I should be looking at each line of css? It's difficult to notice those types of changes in zeplin.

Would it be difficult to make it cover the whole screen? No. Would it be difficult to make it a separate screen just for narrow screens? A bit. Do you have designs for this? Regardless, full screen modal or whatever it becomes could be done as a separate PR.

virtuacoplenny:
bold share link, .8 opacity modal
Screen Shot 2019-06-17 at 2 41 41 PM

PR #375, opened 2019-06-17
Removes XMPP MUC password pairing strategy

paweldomas:


spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -119,7 +116,7 @@ export class BaseRemoteControlService extends Emitter {
      * @returns {string}
      */
     getJoinCode() {
-        throw new Error();
+        return this._options.joinCode;
```
virtuacoplenny: Will `this._options.joinCode` always be valid join code? It doesn't expire?

PR #380, opened 2019-06-19
Backend calendar fixes

paweldomas:


spot-client/src/spot-tv/ui/loaders/withCalendar.js
```diff
@@ -104,13 +114,23 @@ export class CalendarLoader extends AbstractLoader {
...
+     *
+     * @returns {void}
+     */
+    maybeStartPolling() {
+        if (this.props.isSetupComplete && this.state.loaded && !this.state.isPollingStarted) {
+            calendarService.startPollingForEvents({
+                email: this.props.calendarEmail,
+                jwt: this.props.jwt
             });
+            this.setState({ isPollingStarted: true });
```
virtuacoplenny: Do you need this flag? What if the service could no-op if already polling?

paweldomas: yeah it is already doing no-op

PR #384, opened 2019-06-20
feat(controller): provide way to go back to join code

virtuacoplenny:
Screen Shot 2019-06-20 at 9 35 51 AM

### DIFF
--- a/spot-controller/App.js
+++ b/spot-controller/App.js
@@ -28,10 +28,14 @@ export default class App extends React.Component {
 
         this.state = {
             loading: __DEV__,
-            remoteControlUrl: __DEV__ ? null : DEFAULT_URL
+            remoteControlUrl: __DEV__ ? null : DEFAULT_URL,
+            webViewKey: 0
         };
 
+        this._sideMenuRef = React.createRef();
+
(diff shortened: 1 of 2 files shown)


spot-controller/App.js
```diff
@@ -127,7 +132,20 @@ export default class App extends React.Component {
...
         return (
-            <SettingsMenu onClearRemoteUrl = { this._onClearRemoteUrl } />
+            <SettingsMenu
+                onClearRemoteUrl = { this._onClearRemoteUrl }
+                onResetApp = { this._onResetApp } />
         );
     }
+
+    /**
+     * Ah shit.
```
paweldomas: :)

paweldomas: lol

virtuacoplenny: This is what my code looks like without its makeup.

virtuacoplenny:
> LGTM why is this webview key needs to be changed ?

Looks like I lost my comment about it, which I will add back. I couldn't find a decent way to make the webview go back to the initial URL. So instead I force a re-render by changing the key.

PR #386, opened 2019-06-20
feat(updates): auto-reload once a day for backend flow

virtuacoplenny:


spot-client/src/common/app-state/config/default-config.js
```diff
@@ -62,6 +62,15 @@ export default {
         TRANSMISSION_DELAY: undefined
     },
 
+    UPDATES: {
+        START_HOUR: typeof process.env.UPDATE_START_HOUR === 'undefined'
+            ? 2
+            : process.env.UPDATE_START_HOUR,
+        END_HOUR: typeof process.env.UPDATE_END_HOUR === 'undefined'
+            ? 20
```
paweldomas: why `20` by default ? shouldn't it be night hours ? from 2 - 5 or 2 - 4 ?

spot-client/src/common/auto-update/AutoUpdateChecker.js
```diff
@@ -0,0 +1,136 @@
...
+import { connect } from 'react-redux';
+
+import { getUpdateEndHour, getUpdateStartHour } from 'common/app-state';
+import { logger } from 'common/logger';
+import { date } from 'common/utils';
+
+
+const checkInterval = 5 * 60 * 1000; // 5 minutes in milliseconds
+
+const lastLoadTime = new Date();
```
paweldomas: somehow this does not feel very accurate because it's inside of a component which will be refreshed on navigation between views ? what we care about are page loads not component loads

paweldomas: As you know Im not a big fan of smart component, but even if I try to like it then somehow this feels to fit more into a service which is not tied to a lifetime of a component (because of the issue mentioned in the comment above).

paweldomas: I'm thinking that a component is fine to be plugged into the UI screen that supports the reload, but maybe it could work together with a service which will be accurate about tracking the last load times ?

paweldomas: hmm wait I'm just stupid. This date will be when the page was loaded, because it is initialized when the code is loaded, right ?

spot-client/src/common/app-state/config/default-config.js
virtuacoplenny: I think I rushed this PR a bit more than usual. The 20 was set for debugging.

spot-client/src/common/auto-update/AutoUpdateChecker.js
virtuacoplenny: Right, that's why; it's essentially when the js bundle has been executed by the browser. I was thinking I'd make a model/service for this but forgot too. The modal can probably keep track of polling and use pub/sub, that way the little logic there is can be moved around a bit. I'll do that.

paweldomas: No, wait. This is good for now.

virtuacoplenny:
Rebased onto latest master to fix conflicts.

PR #388, opened 2019-06-21
fix(meeting): make sure meeting fills screen

virtuacoplenny:
Works around an issue currently in beta where
100% height on a child of a parent with 100vh
does not cause the child to also be 100vh.
Also, using vh for the child is more consistent.


virtuacoplenny:
I'll keep trying to make a bare repro against beta. Until then, this will at least work around the issue.

PR #391, opened 2019-06-21
ref(Spot Remote/backend): do not disconnect when no Spot TV in MUC

paweldomas:
In the backend mode there's no need to drop the RCS connection when Spot TV is not in the MUC.

### DIFF
--- a/spot-client/src/common/remote-control/remoteControlClient.js
+++ b/spot-client/src/common/remote-control/remoteControlClient.js
@@ -431,7 +431,20 @@ export class RemoteControlClient extends BaseRemoteControlService {
 
             if (this._getSpotId() === from) {
                 logger.log('Spot TV left the MUC');
-                this._onDisconnect(CONNECTION_EVENTS.SERVER_DISCONNECTED);
+                if (this._options.backend) {
+                    // With backend it is okay for remote to sit in the MUC without Spot TV connected.
+                    this._lastSpotState = {
+                        spotId: undefined
+                    };
+                    this.emit(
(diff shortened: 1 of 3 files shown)


spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -431,7 +431,20 @@ export class RemoteControlClient extends BaseRemoteControlService {
...
+                        spotId: undefined
+                    };
+                    this.emit(
+                        SERVICE_UPDATES.SERVER_STATE_CHANGE,
+                        {
+                            updatedState: this._lastSpotState
+                        }
+                    );
+                } else {
+                    this._onDisconnect(CONNECTION_EVENTS.SERVER_DISCONNECTED);
```
virtuacoplenny: What would it look like to have the listener of CONNECTION_EVENTS.SERVER_DISCONNECTED decide whether to disconnect or wait?

paweldomas: I think that would require another type of event because SERVER_DISCONNECTED is currently treated as RCS down.

paweldomas: Do you mean the listener that is in the action ? This way it will have to be aware of the backend vs no backend deal vs right now it is kind of hidden behind the RCS layer.

paweldomas: Also it would be a problem for those components that are checking the `isConnectedToSpot` condition. They would have to handle this case differently based on the backend vs no backend case.

PR #398, opened 2019-06-25
fix(setup): prevent selecting same camera for screenshare

virtuacoplenny:
Do so by removing the  selected camera from the options
for screenshare dongle, but show all available cameras
in the camera dropdown. Also unselect the screenshare
dongle if it matches the selected camera.

Jun-24-2019 17-29-17


spot-client/src/spot-tv/ui/components/setup/select-media/select-media.js
```diff
@@ -22,6 +22,15 @@ import SpeakerPreview from './speaker-preview';
 
 import { wiredScreenshareService } from './../../../../wired-screenshare-service';
 
+/**
+ * @typedef {Object} Device
```
paweldomas: ❤️

PR #402, opened 2019-06-30
fix(backend): use jwt only for configured domains

virtuacoplenny:


spot-client/src/common/app-state/config/default-config.js
```diff
@@ -155,7 +157,26 @@ export default {
...
+        /**
+         * A list of known which the JWT, provided by the pairing service,
+         * is valid for and should be used with when joining meetings.
+         *
+         * @type {Array<string>}
+         */
+        jwtDomains: [],
+
+        /**
+         * The backend URL for registering a Spot-TV.
```
paweldomas: So to be precise the pairing service is used by both Spot TV and Spot Remote in order to exchange a pairing code for an access token.

virtuacoplenny:
Force pushed to update jsdoc for config.SPOT_SERVICES.pairingServiceUrl.

PR #403, opened 2019-07-01
ref: dynamically load external scripts

virtuacoplenny:


virtuacoplenny:
I think it's best to make it dynamic, assuming there are no bugs, because then it wont be loaded when it doesn't have to be; it should only be loaded when spot-tv has the open-source flow and has google integration enabled. This prevents spot-remotes from loading gapi when they don't have to. gapi used to be hardcoded because, in the early days of spot, only the open-source flow was required and only google calendar.

PR #410, opened 2019-07-04
fix(backend): prevent infinite reconnects when backend errors

virtuacoplenny:
@paweldomas, is this too aggressive?


spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -94,6 +94,17 @@ export class SpotBackendService {
         return Boolean(this.registration && this.registration.refreshToken);
     }
 
+    /**
+     * Checks if the error is one that is fatal and should cannot be retried.
```
paweldomas: a typo ?

PR #417, opened 2019-07-11
feat(logging): show device id in admin debug

virtuacoplenny:
Don't know where I'll shove it for spot-remote.

Screen Shot 2019-07-11 at 3 00 15 PM


virtuacoplenny:
Maybe we won't really need this if there will be feedback submission at one point.

PR #426, opened 2019-07-16
Datachannel signaling

paweldomas:
Opening a draft so we can discuss things here

### DIFF
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -444,7 +444,6 @@ export class BaseRemoteControlService extends Emitter {
         this._processMessage(messageType, from, data);
     }
 
-    /* eslint-disable no-empty-function */
     /**
      * Method called when a remote control command is received over the P2P channel.
      *
@@ -456,8 +455,8 @@ export class BaseRemoteControlService extends Emitter {
      * @returns {void}
      */
     _onP2PRemoteControlMessageReceived() {
+        // Implementation is optional. At the time of this writing only the server implements this method.
     }
-    /* eslint-enable no-empty-function */
 
     /**
      * Callback invoked when Spot-TV or a Spot-Remote has a status update.
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -170,9 +170,6 @@ export class BaseRemoteControlService extends Emitter {
                     MESSAGES.REMOTE_CONTROL_P2P,
                     data
                 ).catch(error => logger.error('Failed to send p2p message', { error }));
-            },
-            onRemoteControlMessageReceived: data => {
-                this._onP2PRemoteControlMessageReceived(data);
             }
         }, {
             iceServers: this.xmppConnection.getJitsiConnection().xmpp.connection.jingle.p2pIceConfig.iceServers
@@ -415,20 +412,6 @@ export class BaseRemoteControlService extends Emitter {
         logger.log('RemoteControlService received message', { messageType });
 
         this._processMessage(messageType, from, data);
-    }
-
-    /**
-     * Method called when a remote control command is received over the P2P channel.
-     *
-     * @param {string} remoteAddress - The remote address to which an ack needs to be sent to.
-     * @param {number} requestId - The request ID to be used for an ack.
-     * @param {string} command - A remote control service command.
-     * @param {Object} data - Any command specific extra data(if any).
-     * @protected
-     * @returns {void}
-     */
-    _onP2PRemoteControlMessageReceived() {
-        // Implementation is optional. At the time of this writing only the server implements this method.
     }
 
     /**
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -498,14 +498,10 @@ export class BaseRemoteControlService extends Emitter {
      *
      * @param {string} from - The remote address from which the message has been received.
      * @param {string} data - Any message data.
-     * @private
+     * @protected
      * @returns {void}
      */
     _processRemoteControlP2PMessage({ from, data }) {
-        if (!this._p2pSignaling) {
-            this._createP2PSignalingConnection(/* a server type of connection */ true);
-        }
-
         if (this._p2pSignaling) {
             this._p2pSignaling.processP2PMessage({
                 data,
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -173,32 +173,11 @@ export class BaseRemoteControlService extends Emitter {
             },
             onRemoteControlMessageReceived: data => {
                 this._onP2PRemoteControlMessageReceived(data);
-            },
-            onStatusUpdateReceived: (remoteAddress, newStatus) => {
-                logger.log('On Spot TV status received P2P', {
-                    remoteAddress,
-                    newStatus
-                });
-                this._onSpotTvStatusReceived(remoteAddress, newStatus);
             }
         }, {
             iceServers: this.xmppConnection.getJitsiConnection().xmpp.connection.jingle.p2pIceConfig.iceServers
         });
     }
-
-    /* eslint-disable no-unused-vars */
-    /**
-     * FIXME.
-     *
-     * @param {string} from - FIXME.
-     * @param {Object} state - FIXME.
-     * @private
-     * @returns {void}
-     */
-    _onSpotTvStatusReceived(from, state) {
-        // This is to be implemented only by the client
-    }
-    /* eslint-enable no-unused-vars */
 
     /**
      * Returns a Promise which is to be resolved/rejected when the initial connection process is
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -279,11 +279,6 @@ export class BaseRemoteControlService extends Emitter {
         if (!this._disconnecting) {
             this._disconnecting = true;
 
-            if (this._p2pSignaling) {
-                this._p2pSignaling.stop();
-                this._p2pSignaling = null;
-            }
-
             this.disconnect()
                 .then(() => this.emit(
                     SERVICE_UPDATES.UNRECOVERABLE_DISCONNECT, reason));
@@ -305,6 +300,11 @@ export class BaseRemoteControlService extends Emitter {
                 backend.constructor.REGISTRATION_UPDATED,
                 this._onBackendRegistrationUpdated
             );
+        }
+
+        if (this._p2pSignaling) {
+            this._p2pSignaling.stop();
+            this._p2pSignaling = null;
         }
 
         const destroyPromise = this.xmppConnection
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -179,11 +179,32 @@ export class BaseRemoteControlService extends Emitter {
             },
             onRemoteControlMessageReceived: data => {
                 this._onP2PRemoteControlMessageReceived(data);
+            },
+            onStatusUpdateReceived: (remoteAddress, newStatus) => {
+                logger.log('On Spot TV status received P2P', {
+                    remoteAddress,
+                    newStatus
+                });
+                this._onSpotTvStatusReceived(remoteAddress, newStatus);
             }
         }, {
             iceServers: this.xmppConnection.getJitsiConnection().xmpp.connection.jingle.p2pIceConfig.iceServers
         });
     }
+
+    /* eslint-disable no-unused-vars */
+    /**
+     * FIXME.
+     *
+     * @param {string} from - FIXME.
+     * @param {Object} state - FIXME.
+     * @private
+     * @returns {void}
+     */
+    _onSpotTvStatusReceived(from, state) {
+        // This is to be implemented only by the client
+    }
+    /* eslint-enable no-unused-vars */
 
     /**
      * Returns a Promise which is to be resolved/rejected when the initial connection process is
--- a/spot-client/src/common/remote-control/BaseRemoteControlService.js
+++ b/spot-client/src/common/remote-control/BaseRemoteControlService.js
@@ -4,9 +4,12 @@ import { generate8Characters } from 'common/utils';
 
 import {
     CLIENT_TYPES,
+    MESSAGES,
     SERVICE_UPDATES
 } from './constants';
 import XmppConnection from './xmpp-connection';
+import P2PSignalingClient from './P2PSignalingClient';
+import P2PSignalingServer from './P2PSignalingServer';
 
 /**
  * The interface for interacting with the XMPP service which powers the
@@ -35,6 +38,15 @@ export class BaseRemoteControlService extends Emitter {
                     .catch(() => { /* swallow unload errors from bubbling up */ });
             }
         );
+
+        /**
+         * A peer-to-peer direct signaling channel. Once established will be used to send/receive remote control
+         * commands.
+         *
+         * @type {P2PSignalingClient|P2PSignalingServer|null}
+         * @protected
+         */
+        this._p2pSignaling = null;
     }
 
     /**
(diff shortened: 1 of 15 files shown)


spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -392,11 +460,40 @@ export class BaseRemoteControlService extends Emitter {
...
      */
-    _processMessage() {
-        return;
+    _processRemoteControlP2PMessage({ from, data }) {
+        if (!this._p2pSignaling) {
+            this._createP2PSignalingConnection(/* a server type of connection */ true);
+        }
+
+        if (this._p2pSignaling) {
+            this._p2pSignaling.processP2PMessage({
```
virtuacoplenny: What happens if the p2p signaling connection isn't ready after it was created a few lines above?

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,336 @@
...
+     * Creates and initializes new {@link PeerConnection} instance for given remote address.
+     *
+     * @param {string} remoteAddress - The remote address string.
+     * @private
+     * @returns {PeerConnection}
+     * @throws an error if {@link PeerConnection} already exists for the given address.
+     */
+    _initPeerConnection(remoteAddress) {
+        if (this._peerConnections.get(remoteAddress)) {
+            throw new Error(`P2P signaling connection already exists for: ${remoteAddress}`);
```
virtuacoplenny: I wonder if it should error. Is there a need to make it known this redundant create call happened?

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,336 @@
...
+                this._onDataChannelMessage(pc.remoteAddress, data);
+            });
+        peerConnection.addListener(
+            PeerConnection.ON_ICE_CANDIDATE, (pc, candidate) => {
+                // No 'candidate' means end of candidates
+                candidate && this._callbacks.onSendP2PMessage(
+                        pc.remoteAddress, { candidate: JSON.stringify(candidate) });
+            });
+
+        return peerConnection;
```
virtuacoplenny: Do you know if this works okay across all supported browsers? I'm assuming so, at this point.

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,336 @@
...
+    }
+
+    /**
+     * Checks if the first {@link PeerConnection} is active.
+     *
+     * @returns {boolean}
+     */
+    isReady() {
+        const first = this.getFirstPeerConnection();
```
virtuacoplenny: Why check the first specifically?

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -62,8 +62,7 @@ export class RemoteControlClient extends BaseRemoteControlService {
      * @returns {Promise} Resolves if the command has been acknowledged.
      */
     adjustVolume(direction) {
-        return this.xmppConnection.sendCommand(
-            this._getSpotId(), COMMANDS.ADJUST_VOLUME, { direction });
+        return this._sendCommand(COMMANDS.ADJUST_VOLUME, { direction });
```
virtuacoplenny: This change could go in on its own? Just trying to figure out stuff we can get in earlier to make the diff clearer.

spot-client/src/common/remote-control/BaseRemoteControlService.js
paweldomas: It is not ready. Those messages here are to exchange offer/answer/ice candidates. The connection is ready only after ice connects. This part of the flow is really server specific where it will start the instance when first message arrives which is most likely an invite.

spot-client/src/common/remote-control/P2PSignalingBase.js
paweldomas: This is a case that is supposed to "never happen". I can log an error and early return from the method instead, but weird things could happen.

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: yes

spot-client/src/common/webrtc/PeerConnection.js
```diff
@@ -0,0 +1,231 @@
...
+    }
+
+    /**
+     * Creates an offer and returns a structure that needs to be transferred over to the remote peer and set using
+     * {@link setOffer}.
+     *
+     * @returns {Promise<RTCSessionDescriptionInit>}
+     */
+    createOffer() {
+        this._initPeerConnection();
```
virtuacoplenny: What do you think about creating the peer connection in the constructor? That way more of the initialization happens while other instance variables are getting set. That would also prevent the case of calling createOffer while a peer connection already exists.

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,364 @@
...
+ * {@link https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer/urls} for more details about the format.
+ */
+/**
+ * A peer to peer(direct) signaling channel used to send/receive remote control commands.
+ */
+export default class P2PSignalingBase {
+    /**
+     * Creates new instance.
+     *
+     * @param {P2PSignalingCallbacks} callbacks - The callbacks structure, see type description for more info.
```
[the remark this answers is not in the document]
paweldomas: I think I was refactoring code passed on LJM proxy service so it just stayed this way. Im having hard time to decide which one should be used. But callbacks seem to make sense when implementing them is mandatory where events could be optional and allow to subscribe many listeners.

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,364 @@
...
+    /**
+     * FIXME.
+     *
+     * @param {string} remoteAddress - The remote address that sent the message.
+     * @param {Object} msg - FIXME.
+     * @protected
+     * @returns {void}
+     */
+    _processDataChannelMessage(remoteAddress, msg) {
+        logger.warn('_processDataChannelMessage ignoring a msg', {
```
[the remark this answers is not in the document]
paweldomas: A fallback for handling commands not handled neither by server nor client impl which leaves doors open for having a place for common processing.

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,364 @@
...
+    /**
+     * Tries to parse given JSON string and convert it into an {@code Object}. If fails logs an error and returns
+     * {@code undefined}.
+     *
+     * @param {string} jsonText - The text to be parsed.
+     * @param {string} errorMessage - A text to be added at the end of the error message to be logged on an error.
+     * @returns {undefined|Object}
+     * @private
+     */
+    _parseJSONorLogError(jsonText, errorMessage) {
```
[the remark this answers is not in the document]
paweldomas: why not? do you mean to swallow it or not catch it and let it happen?

spot-client/src/common/remote-control/P2PSignalingClient.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * Sends remote control command over the underlying P2P data channel.
+     *
+     * @param {string} command - An RCS command to be sent.
+     * @param {string} data - Any data command specific.
+     * @returns {Promise<void>} - Resolved when an ack is received from the server or rejected if anything goes wrong.
+     */
+    sendCommand(command, data) {
```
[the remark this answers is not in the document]
paweldomas: yes, I was thinking similar

spot-client/src/common/remote-control/remoteControlServer.js
```diff
@@ -225,7 +225,13 @@ export class RemoteControlServer extends BaseRemoteControlService {
             return;
         }
 
+        newStatus.timestamp = Date.now();
+
         this.xmppConnection.updateStatus(newStatus);
+
+        if (this._p2pSignaling && this._p2pSignaling.isReady()) {
+            this._p2pSignaling.updateStatus(newStatus);
```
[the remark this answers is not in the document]
paweldomas: yeah, I can hide the readiness check inside `updateStatus`(if that's what you mean), especially that nothing depends on it, it's optional.

spot-client/src/common/webrtc/PeerConnection.js
```diff
@@ -0,0 +1,231 @@
...
+     * @param {RTCIceServer[]} [iceServers] - An array with ICE STUN/TURN servers description as defined by WebRTC. See
+     * {@link https://developer.mozilla.org/en-US/docs/Web/API/RTCIceServer/urls} for more details about the format.
+     */
+    constructor(remoteAddress, iceServers) {
+        super();
+        this.remoteAddress = remoteAddress;
+        this._iceServers = iceServers || DEFAULT_STUN_SERVERS;
+        this._peerConnection = null;
+        this._dataChannel = null;
+        this._currentStatus = false;
```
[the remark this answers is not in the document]
paweldomas: yeah that naming is messed up

spot-client/src/common/webrtc/PeerConnection.js
paweldomas: I will look into that

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -210,6 +257,12 @@ export class BaseRemoteControlService extends Emitter {
         // If XMPP server is not working calling disconnect triggers onDisconnect again.
         if (!this._disconnecting) {
             this._disconnecting = true;
+
+            if (this._p2pSignaling) {
```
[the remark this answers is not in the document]
paweldomas: yes

spot-client/src/common/remote-control/BaseRemoteControlService.js
paweldomas: This code is not only server specific. Both P2P client and server need to exchange either offer or answer and both process ice candidates.

spot-client/src/common/remote-control/remoteControlServer.js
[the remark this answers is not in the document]
paweldomas: Unfortunately it can not be moved there as it depends on the `xmppConnection` field being initialized.

spot-client/src/common/remote-control/BaseRemoteControlService.js
paweldomas: Okay so I misunderstood and the part above which creates the connection is server specific, but the part where a message is being processed is not.

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,364 @@
...
+        if (!peerConnection) {
+            logger.error('processMessage - no PeerConnection', {
+                data,
+                from
+            });
+
+            return;
+        }
+
+        if (peerConnection.remoteAddress !== from) {
```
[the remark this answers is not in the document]
paweldomas: It is supposed to never happen, but it's here to catch bugs.

spot-client/src/common/remote-control/P2PSignalingClient.js
```diff
@@ -0,0 +1,129 @@
...
+                commandPromise.resolve();
+            } else {
+                logger.warn('No P2P command promise exists', { requestId });
+            }
+
+            this._p2pSignalingRequests.delete(requestId);
+        } else if (msg.command === 'status' && msg.newStatus) {
+            this._callbacks.onStatusUpdateReceived(remoteAddress, msg.newStatus);
+        } else {
+            super._processDataChannelMessage(remoteAddress, msg);
```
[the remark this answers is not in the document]
paweldomas: It doesn't, but it logs a warning which would be duplicated in both server and the client. Also is a place where you would put common msg processing in future. Because it's in future I can get rid of it if you insist.

spot-client/src/common/remote-control/P2PSignalingClient.js
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
```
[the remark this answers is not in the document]
paweldomas: This comment is displayed by Github out of context, but I suppose you're asking what do we need the XMPP connection for in P2P signaling. So two things:
- for getting ICE servers for LJM connection
- for sending signaling msgs, but this one can probably act conditionally if the xmppConnection exists, so is not really a blocker here.

spot-client/src/common/remote-control/P2PSignalingBase.js
paweldomas: It was not working on webview, but seems to be okay now

spot-client/src/common/remote-control/P2PSignalingBase.js
paweldomas: It's been moved to the client version which only cares to have one PeerConnection.

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -138,6 +150,62 @@ export class BaseRemoteControlService extends Emitter {
...
+     * commands.
+     *
+     * @param {boolean} isServer - Whether to create a server or a client P2P signaling connection.
+     * @protected
+     * @returns {void}
+     */
+    _createP2PSignalingConnection(isServer) {
+        logger.log('Will try to establish P2P signaling channel', { isServer });
+
+        const P2PSignalingType = isServer ? P2PSignalingServer : P2PSignalingClient;
```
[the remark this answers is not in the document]
paweldomas: I don't have clear idea on this one, because I don't know how to get rid of duplication in case each subclass is creating it. Would `getP2PSignalingType` abstract method be okay with you?

spot-client/src/common/remote-control/P2PSignalingClient.js
```diff
@@ -0,0 +1,156 @@
...
+        });
+
+        const peerConnection = this._getFirstPeerConnection();
+
+        if (!peerConnection) {
+            throw new Error('No PeerConnection');
+        }
+
+        // FIXME time may not be 100% reliable for unique IDs
+        const requestId = new Date().getTime();
```
virtuacoplenny: Why not use one of the other helpers in common for creating unique ids? Could it even be an increment or a combination of some id of the remote and an increment?

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -97,6 +98,23 @@ export class RemoteControlClient extends BaseRemoteControlService {
...
+     * commands.
+     *
+     * @param {boolean} isServer - Whether to create a server or a client P2P signaling connection.
+     * @protected
+     * @returns {void}
+     */
+    _createP2PSignalingConnection(isServer) {
+        super._createP2PSignalingConnection(isServer);
+
+        !isServer
```
virtuacoplenny: When is remoteControlClient going to get isServer true?

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -501,6 +549,13 @@ export class RemoteControlClient extends BaseRemoteControlService {
             this._waitForSpotTvTimeout = null;
         }
 
+        if (!this._p2pSignaling) {
+            this._createP2PSignalingConnection(false);
```
virtuacoplenny: Why create the connection if signaling is not enabled?

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -487,6 +513,28 @@ export class RemoteControlClient extends BaseRemoteControlService {
...
+     * @private
+     * @returns {void}
+     */
+    _onSpotTvStatusReceived(from, state) {
+        // NOTE There will be no timestamp in the initial Spot TV presence
+        const { timestamp } = state;
+
+        const currentTimestamp = this._lastSpotState && this._lastSpotState.timestamp;
+
+        if (currentTimestamp && timestamp && currentTimestamp >= timestamp) {
```
virtuacoplenny: Something converts the timestamp to a number, correct? I'm thinking about the case where presence is in xml, not json, and the timestamp is somehow obtained from it.

spot-client/src/common/remote-control/P2PSignalingClient.js
```diff
@@ -0,0 +1,156 @@
...
+
+            count && logger.log('Rejecting signaling promises', { count });
+
+            for (const promise of this._p2pSignalingRequests.values()) {
+                promise.reject('P2P signaling channel closed');
+            }
+
+            this._p2pSignalingRequests.clear();
+        }
+    }
```
virtuacoplenny: Is there a mechanism built in to try to recreate the data-channel or peer connection if either become closed?

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -0,0 +1,353 @@
...
+            return;
+        }
+
+        const offer = this._parseJSONorLogError(data.offer, 'offer message');
+
+        if (!offer) {
+            return;
+        }
+
+        // Try catch block in case something's wrong with WebRTC version or not supported
```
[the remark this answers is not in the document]
paweldomas: When I was testing webview it was throwing errors because of incompatible API and I found that catching the error here prevents from breaking the app and makes P2P optional. After fix currently it's working everywhere where I tested it, but that does not cover older browsers.

spot-client/src/common/remote-control/P2PSignalingClient.js
paweldomas: Right, I will do that

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: In theory never, but you never know :P Just trying to follow the api.

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: JSON.parse coverts it on the presence to XML conversion. I don't understand the XML case, when could it happen?

spot-client/src/common/remote-control/P2PSignalingClient.js
paweldomas: Currently only if either TV or remote reconnects.

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: Because then I have to do the flag logic in two places, but maybe it's better to just not create it. I will change that, because the current behavior seems weird.

spot-client/src/common/remote-control/P2PSignalingClient.js
paweldomas: But I think this is good enough for the start. The reconnect logic is on the higher level which means whenever network connectivity is lost p2p will be restarted as well. I'd say we add analytics to see when remote is connected over p2p and the figure it out from there if it turns out there are corner cases worth chasing.

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: It will look better when client and server create the type then this flag will be gone

PR #427, opened 2019-07-16
fix(backend): notify when jwt is refreshed

virtuacoplenny:


spot-client/src/common/backend/SpotBackendService.test.js
```diff
@@ -44,7 +44,6 @@ describe('SpotBackendService', () => {
 
             it('refreshes the token automatically', () => {
                 jest.useFakeTimers();
-                fetch.mockResponseOnce(JSON.stringify(MOCK_RESPONSE));
```
virtuacoplenny: Something weird is going on with fetch mock. It's almost like it's not clearing state between runs.

PR #431, opened 2019-07-19
Some reducer tests

virtuacoplenny:


spot-client/src/common/app-state/setup/reducer.test.js
```diff
@@ -0,0 +1,75 @@
...
+            displayName
+        }))
+        .toEqual(expect.objectContaining({ displayName }));
+    });
+
+    it('sets a jwt', () => {
+        const jwt = 'test-jwt';
+
+        expect(setupReducer(undefined, {
+            type: types.SET_JWT,
```
paweldomas: I'm too lazy to look at the code, but don't we have action functions ? Would it not be better to use that actions instead ? This duplicates the code a little bit.

virtuacoplenny: It's weird to me for a unit test but also makes sense to test together to hide implementation details. action creators could be used to fire reducers and then selectors used. This might require initializing a redux store with each test, but maybe that's okay.

virtuacoplenny: I'm converting these to integration tests.

spot-client/src/common/app-state/modal/modal.test.js
```diff
@@ -0,0 +1,70 @@
...
+        ({ dispatch, getState } = createStore(combineReducers({ modal: modalReducer })));
+    });
+
+    it('by default has no modal', () => {
+        verifyNoModal();
+    });
+
+    describe('showing modals', () => {
+        it('sets the modal component to display with its props', () => {
+            const mockProps = {};
```
paweldomas: how about putting something here ? just thinking an empty object is not ideal here

virtuacoplenny: Added a key-value pair into the props. I did a fixup in the last commit and rebased onto latest master while I was at it.

PR #436, opened 2019-07-24
feat(backend): add an id to requests

virtuacoplenny:


spot-client/setupTests.js
```diff
@@ -6,3 +7,7 @@ Enzyme.configure({ adapter: new EnzymeAdapter() });
 jest.mock('./src/common/logger');
 
 global.fetch = require('jest-fetch-mock');
+
+global.crypto = {
+    getRandomValues: buffer => crypto.randomFillSync(buffer)
```
paweldomas: why this change ?

paweldomas: just not sure how is it related to adding `request-id` header

virtuacoplenny: guid creation. I could mock out the guid method itself, if that's preferred.

PR #448, opened 2019-07-29
Clear stored access token on request error

paweldomas:


spot-client/src/common/backend/SpotBackendService.test.js
```diff
@@ -82,6 +82,21 @@ describe('SpotBackendService', () => {
                     })
                     .then(() => expect(onUpdateCallback).toHaveBeenCalled());
             });
+
+            it('clears the registration on room info error', () => {
+                spotBackendService.register(MOCK_PAIRING_CODE).then(() => {
```
virtuacoplenny: You need to return the promise when needing to wait for async to finish or else the test will not wait.

PR #453, opened 2019-07-30
fix(BaseRemoteControlService): infinite loop on rejected promise

paweldomas:
Both remote control server and remote control client further extend the connect promise. The 'catch' must be attached on top of the promises created by the subclasses in order to always reset
the 'xmppConnectionPromise'. Otherwise the app will loop forever on the rejected promise.


paweldomas:
Merged with https://github.com/jitsi/jitsi-meet-spot/pull/454

PR #457, opened 2019-07-31
fix(controller): stop targeting iOS 10

virtuacoplenny:
I can't get confirmation on what versions to support so I'm picking the one that starts supporting caret-color. I leave it to you to keep me in check.


saghul:
Any reason why 11.3 was chosen as the target instead of 11.0? LGTM otherwise.

virtuacoplenny:
According to caniuse, the css property caret-color started being supported on ios 11.3. Without being able to get an answer on what the minimum support should be, I went with what was convenient for me to fix a really minor bug; the bug being a hidden input having a visible cursor on ios.

saghul:
👍

saghul:
Please wait for Zoli before merge.

PR #461, opened 2019-08-01
feat(spot-remote): dispatch SET_ROOM_ID

paweldomas:
This will also adjust the logging id based on the room ID.


virtuacoplenny:
But do you really want TVs and and Remotes to share the same id?

virtuacoplenny:
So Pawel showed me what I should have remembered and there's middleware that adds a room id suffix to differentiate tvs and remotes.

PR #462, opened 2019-08-01
feat(backend): get a long lived pairing token

virtuacoplenny:
Screen Shot 2019-08-01 at 2 31 03 PM
Screen Shot 2019-08-01 at 2 31 00 PM


spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
```diff
@@ -108,7 +111,8 @@ function mapDispatchToProps(dispatch) {
             return dispatch(createSpotTVRemoteControlConnection({
                 pairingCode,
                 retry: false
-            }));
+            }))
+            .then(() => dispatch(generateLongLivedPairingCode()));
```
paweldomas: There's one problem with doing that discovered while testing. If the `generateLongLivedPairingCode()` fails then the app gets stuck on the screen because the first step of creating the connection was completed and trying again it says that the connection still exists.

So one fix would be to disconnect when the `dispatch(generateLongLivedPairingCode())` fails. Another one would be to take fetching of the long lived code onto the next screen with the option to retry if it fails. I see the second option as a possible follow up PR, but we could start with the first one just to have it in the app already.

paweldomas:
Still I like the PR just that one detail. Thanks for taking care of this !

paweldomas:
I have tried to fix that disconnected in https://github.com/jitsi/jitsi-meet-spot/pull/463 plus added logging and spot-admin feature to support that request.

PR #464, opened 2019-08-02
feat(webdriver): allow running tests with backend

virtuacoplenny:


spot-webdriver/specs/in-meeting.spec.js
```diff
@@ -14,6 +14,10 @@ describe('While in a meeting ', () => {
             .waitForVisible();
     });
 
+    afterEach(() => {
+        spotSession.resetConnection();
```
paweldomas: Is there any test where we would not want this to happen after a run ? Maybe I wish this was more automatic/implicit/happening behind the scenes, but I understand if not doable or a lot of effort.

virtuacoplenny: I might be able to figure out a way but it might have to be in webdriverio. Before I had a global afterEach that ran after each test where I could include global cleanup. With the latest update each file runs in its own instance, so that global afterEach stopped working. What could be done is in the webdriverio config a cleanup call could be done, but then the refactoring would be to have a factory function for the SpotSession so there could be a global/centra place that holds onto all session for the cleanup call to access (instead of an instance scoped to the test).

In general we want a cleanup call though because of the spot-tv connection conflicts caused by jids. It's not that it's a lot of effort to workaround it's that this current impl was the easiest for now. I could follow up tomorrow and try the global cleanup in the webdriverio config but also I had some trouble with it respecting async.

virtuacoplenny: Rebased and tried creating a global session store so that the webdriverio config can clear connections in its afterTest hook.

PR #468, opened 2019-08-05
Add legal links

virtuacoplenny:
I'm uncertain where I'm actually supposed to put these links. At first I was told to put them on the help page but then told to put them in the settings panel under a cog. So I put the links on the help page and put them in the settings panel, but not under a cog because there really isn't much in the settings panel.

Screen Shot 2019-08-05 at 10 44 20 AM
Screen Shot 2019-08-05 at 9 38 09 AM


spot-client/src/spot-remote/ui/views/help.js
```diff
@@ -67,11 +68,41 @@ class Help extends React.Component {
...
+     * Creates links to legal documents related to using Spot.
+     *
+     * @private
+     * @returns {ReactElement}
+     */
+    _renderLegalLinks() {
+        return (
+            <div className = 'legal-links'>
+                <a
+                    href = 'https://www.8x8.com/terms-and-conditions/privacy-policy'
```
saghul: Can't these be configurable, like on Jitsi Meet?

spot-client/src/spot-remote/ui/views/help.js
```diff
@@ -67,11 +68,41 @@ class Help extends React.Component {
...
+                    rel = 'noopener noreferrer'
+                    target = '_blank'>
+                    Privacy Policy
+                </a>
+                <a
+                    href = 'https://www.8x8.com/terms-and-conditions/8x8-end-user-terms-of-use'
```
saghul: Ditto

saghul:
> So I put the links on the help page and put them in the settings panel, but not under a cog because there really isn't much in the settings panel.

IMHO the cog is needed, there no indication there is a settings menu otherwise.

virtuacoplenny:
> > So I put the links on the help page and put them in the settings panel, but not under a cog because there really isn't much in the settings panel.
> 
> IMHO the cog is needed, there no indication there is a settings menu otherwise.

The request sounded like adding a cog within the settings menu itself. There is an open ticket or two requesting designs for how we should indicate the existence of a settings menu, as Pawel and Zoli and QA have brought this up as well. On my end I haven't prioritized cowboy-ing a solution.

spot-client/src/spot-remote/ui/views/help.js
virtuacoplenny: You think we should?

saghul: The Jitsi "Spot" app would have different links. Yes, they may point to the same place, but we redirect from jitsi.org.

saghul:
> The request sounded like adding a cog within the settings menu itself. There is an open ticket or two requesting designs for how we should indicate the existence of a settings menu, as Pawel and Zoli and QA have brought this up as well. On my end I haven't prioritized cowboy-ing a solution.

A hamburger menu on the top left, like we do in jitsi meet sounds like an easy win maybe?

spot-client/src/spot-remote/ui/views/help.js
virtuacoplenny: I sense you do want is so I'm going to make it configurable in a follow up commit today, probably the first thing I work on

virtuacoplenny:
I guess it's time for me to go off design script again. I'll do it today.

spot-client/src/spot-remote/ui/views/help.js
saghul: Thank you! That shouldn't block this PR though.

virtuacoplenny:
I demoed spot to a couple people and they both said the menu should remain hidden because it's not for users to use but rather for the admin. So I'm back to not doing anything until getting product + design to commit to a design.

PR #478, opened 2019-08-07
Refresh the access token and retry if backend replies with 401

paweldomas:
Adds a logic which will retry a backend request in case it responds with 401.

Still not sure whether the calendar request should be included or not ?

### DIFF
--- a/spot-client/src/spot-tv/backend/SpotTVBackendService.test.js
+++ b/spot-client/src/spot-tv/backend/SpotTVBackendService.test.js
@@ -1,0 +1,135 @@
+import { persistence } from 'common/utils';
+
+import { SpotTvBackendService } from './SpotTvBackendService';
+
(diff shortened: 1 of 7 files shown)


spot-client/src/spot-tv/backend/SpotTVBackendService.test.js
```diff
@@ -0,0 +1,135 @@
+import { persistence } from 'common/utils';
+
+import { SpotTvBackendService } from './SpotTvBackendService';
+
+// FIXME duplicated with SpotBackendService.test.js
```
virtuacoplenny: I think the duplication with these mocks is fine. If you really wanted you could create helpers that do these mocks but I couldn't argue for them at this point, to be honest. I understand the un-easiness with copy/paste but this seems small enough to be okay (admittedly that's how problems start though)

PR #484, opened 2019-08-08
fix(nav): prevent overflow cutoff on safari

virtuacoplenny:
Happens on wide resolutions.

Issue:
Screen Shot 2019-08-08 at 12 37 01 PM

With potential fix:
Screen Shot 2019-08-08 at 12 37 18 PM


virtuacoplenny:
Let's see what this break.

PR #485, opened 2019-08-08
ref(device-id): include device type in device ID

paweldomas:
Will add 'spot-tv', 'remote-perm' or 'remote-temp' as part of the device ID to be able to tell the device type instantly.

### DIFF
--- a/spot-client/src/spot-remote/app-state/actions.js
+++ b/spot-client/src/spot-remote/app-state/actions.js
@@ -5,9 +5,9 @@ import {
     getSpotServicesConfig,
     setCalendarEvents,
     setReconnectState,
-    setRoomId,
     setSpotTVState
 } from 'common/app-state';
+import { setSpotInstanceInfo } from 'common/app-state/device-id';
 import { createAsyncActionWithStates } from 'common/async-actions';
 import { isBackendEnabled, SpotBackendService } from 'common/backend';
 import { history } from 'common/history';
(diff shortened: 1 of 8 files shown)


spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -105,7 +105,11 @@ export function connectToSpotTV(joinCode, shareMode) {
 
             const roomId = roomProfile && roomProfile.id;
 
-            roomId && dispatch(setRoomId(roomId));
+            roomId && dispatch(setSpotInstanceInfo({
+                roomId,
+                isSpotTv: false,
+                isPairingPermanent: backend && backend.isPairingPermanent()
+            }));
```
virtuacoplenny: Is any of this info redundant with something that's already in state? I know that setup.isSpot exists already. Could permanent vs temporary be gleaned in another way? Basically I'm wondering if this plumbing that does not need to exist if existing state can be re-used readily.

paweldomas: Yes, I will look into that. My only problem was that device ID is kind of basic feature which should be as independent as possible, but if it starts importing other features here and there this will not be the case and will be risk of introducing circular dependencies. Right now it's that this info is submitted by the action from a feature that uses it, so it goes top down.

paweldomas: You know unless middlewares would be a separate thing. Then it would be ok I think.

virtuacoplenny: We've brought making middleware separate before. Didn't jitsi-meet mobile also bring it up? From my perspective, we are on this path:
- Services that do its stateful stuff as needed
- Pubsub (or whatever communication mechanism) to store only necessary state in the UI
- Likely shoving logic out of actions and into middleware, although for now we're shoving logic into actions and out of the smart components
- Separate redux state and middleware from the rest of the app instead of doing the feature folder approach of jitsi-meet

PR #486, opened 2019-08-08
feat(rcs): record what type of remotes are connected

virtuacoplenny:
Prep-work for automatically advancing setup when a remote has been connected.


spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -181,7 +181,7 @@ export class BaseRemoteControlService extends Emitter {
             return `${prefix}-${generate8Characters()}`;
         }
 
-        return undefined;
+        return `remote-${generate8Characters()}`;
```
paweldomas: how about doing `remote-perm` only on permanent and do that fallback in any other case including temporary remote ?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -357,6 +357,25 @@ export class BaseRemoteControlService extends Emitter {
...
+     * Helper for determining the type of the Spot client based on its jid.
+     *
+     * @param {string} jid - The jid which contains identifying information
+     * about the client.
+     * @private
+     * @returns {string}
+     */
+    _getTypeFromResource(jid) {
+        if (jid.includes('/spot-tv')) {
+            return 'spot-tv';
```
paweldomas: Can we have constants defined for those types ?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -366,14 +367,14 @@ export class BaseRemoteControlService extends Emitter {
...
-            return 'permanent-remote';
+        if (jid.includes(`/${CLIENT_TYPES.SPOT_TV}`)) {
+            return CLIENT_TYPES.SPOT_TV;
+        } else if (jid.includes(`/${CLIENT_TYPES.SPOT_REMOTE_PERMANENT}`)) {
+            return CLIENT_TYPES.SPOT_REMOTE_PERMANENT;
         }
 
         // Both "remote-temp" and "remote-" are considered temporary.
-        return 'temporary-remote';
+        return CLIENT_TYPES.SPOT_REMOTE_TEMPORARY;
```
paweldomas: Is the comment above(line 376) outdated now ?

PR #491, opened 2019-08-09
ref: move permanent pairing code to common/backend and fix analytics

paweldomas:
Moves permanent pairing code storage and selectors to common/backend. Sends pairing code analytics for both Spot TV and Remote.


spot-client/src/common/backend/analyticsEvents.js
```diff
@@ -0,0 +1,11 @@
...
+    /**
+     * When backend pairing is lost.
+     */
+    PAIRING_FAIL: 'permanent-pairing-code-fail',
+
+    /**
+     * When backend pairing is established.
+     */
+    PAIRING_SUCCESS: 'permanent-pairing-code-success'
+};
```
virtuacoplenny: I thought all analytics events were going to be in files within /analytics directories. Should /backend be concerned with analytics? I see below you added analytics-related middleware to backend as well.

spot-client/src/common/utils/store-peristence.test.js
```diff
@@ -0,0 +1,135 @@
...
+                        permanentPairingCode: SPOT_REMOTE_PERM_CODE
+                    }
+                });
+
+            expect(getPersistedState()).toMatchObject({
+                backend: {
+                    permanentPairingCode: SPOT_TV_PERM_CODE
+                }
+            });
+        });
```
virtuacoplenny: Nice. Maybe we want a test that fails in a few weeks as a reminder to delete the legacy support.

spot-client/src/common/backend/index.js
```diff
@@ -6,3 +8,6 @@ export {
     getRemotePairingCode,
     refreshAccessToken
 } from './utils';
+
+import './middleware';
+import './reducer';
```
virtuacoplenny: Are you sure we want to now have state in common/backend? Do we need it?

paweldomas: We can have it either here or in "spot remote/backend" and "spot tv/backend". I though to keep it in one/common place as it turned out that it's the same thing for both TV and remote.

spot-client/src/common/backend/analyticsEvents.js
paweldomas: I can have this moved to common/analytics

virtuacoplenny: That's what I'm asking about, yes. Is there an explicit reason why you moved away from having all analytics live in /analytics directories?

spot-client/src/common/backend/index.js
virtuacoplenny: My only concern is common/backend having knowledge of redux. If it's necessary for it to do so now then that's what has to happen. Otherwise I would like to keep a separation of redux and the backend logic if, and only if, reasonable.

spot-client/src/common/backend/analyticsEvents.js
paweldomas: No, I will move it there

spot-client/src/common/backend/index.js
paweldomas: Well, really backend service class is not directly involved in redux, but Im thinking of this package as a feature like we have in JitsiMeet. I think it belongs here or I have no better idea. Let me know if you do.

spot-client/src/common/analytics/analyticsEvents.js
```diff
@@ -0,0 +1,11 @@
...
+    /**
+     * When backend pairing is lost.
+     */
+    PAIRING_FAIL: 'permanent-pairing-code-fail',
+
+    /**
+     * When backend pairing is established.
+     */
+    PAIRING_SUCCESS: 'permanent-pairing-code-success'
+};
```
virtuacoplenny: what's wrong with events.js?

paweldomas: Yeah, nothing. Thanks !

PR #493, opened 2019-08-09
feat(controller): stop hiding url changing behind dev

virtuacoplenny:
A masterclass in design.
Screen Shot 2019-08-09 at 2 16 07 PM

### DIFF
--- a/spot-controller/src/setup.js
+++ b/spot-controller/src/setup.js
@@ -1,21 +1,41 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import {
+    Button,
     StyleSheet,
+    Text,
     TextInput,
     View
 } from 'react-native';
+import url from 'url';
+
+import { logger } from './logger';
 
 const styles = StyleSheet.create({
     container: {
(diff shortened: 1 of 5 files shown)


spot-controller/src/setup.js
```diff
@@ -86,6 +116,14 @@ export default class Setup extends React.Component {
      * @returns {void}
      */
     _onSubmitEnteredUrl() {
-        this.props.onSubmitEnteredUrl(this.state.enteredRemoteControlUrl);
+        try {
+            const urlParts = url.parse(this.state.enteredRemoteControlUrl);
+
+            if (/localhost|jitsi.net|8x8.vc/.exec(urlParts.hostname)) {
```
virtuacoplenny: This is the whitelist part. This feels pretty limiting but maybe it's okay for now, meaning the next n-weeks.

virtuacoplenny: Maybe it'd be better to have a lock for this feature? Not sure how we'd go about setting the lock. But I still think this is okay for now, as long as we decide what to do and do it before the next release.

PR #495, opened 2019-08-10
ref(controller): inject js to reset app

virtuacoplenny:
This is an attempt to simply the reset logic.
The drawback is all local storage gets cleared,
including any data that was external to
the redux state being saved, but maybe that is
for the best.

### DIFF
--- a/spot-client/src/index.js
+++ b/spot-client/src/index.js
@@ -28,7 +28,6 @@ import {
     remoteControlClient
 } from 'common/remote-control';
 import {
-    clearPersistedState,
     getPersistedState,
     loadScript,
     setPersistedState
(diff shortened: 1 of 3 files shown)


virtuacoplenny:
Saw a bug with this impl while re-testing, which I'll fix along with the conflicts.

virtuacoplenny:
Force pushed to resolve conflicts for now but there is another change I would like to go in first to fix the bug (onboarding not displaying until relaunch even after reset). Likely there will be conflicts again.

virtuacoplenny:
This should be ready now. The issue with reset not causing onboarding to show should have been fixed as a side effect of #506.

PR #500, opened 2019-08-13
ref: improve logs around the pairing code

paweldomas:


spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -133,9 +133,9 @@ export class SpotBackendService extends Emitter {
      */
     _maybeClearRegistration(error) {
         if (this.isUnrecoverableRequestError(error)) {
+            this.getJwt() && logger.log('Cleared backend registration');
```
virtuacoplenny: Would it be better to always log but include if there was a jwt? Same really with any logs? Always log but include data about the log?

paweldomas: This one about cleared backend registration is very confusing to me when looking through the logs.

PR #503, opened 2019-08-13
feat(RN): report native version as user agent

paweldomas:
Copies the AppInfo module from jitsi-meet and uses the app name, version and build number to construct the WebView's user agent.


spot-controller/src/remote-control.js
```diff
@@ -46,6 +51,11 @@ export default class RemoteControl extends React.PureComponent {
         this._setRef = this._setRef.bind(this);
 
         this._resolveUrl();
+
+        this.userAgentString
+            = __DEV__
+                ? 'SpotController/development'
+                : `${AppInfo.name}/${AppInfo.version}(${AppInfo.buildNumber})`;
```
virtuacoplenny: To make sure the blame is put on me, I requested that this user agent still take into account feature version somehow. The reason for having user agent in the first place was to provide web a way to know what features the react native app supports, in case some backwards compatibility (ie disabling features) is needed.

PR #509, opened 2019-08-16
Fix issues around JWT updated propagation to Redux

paweldomas:


spot-client/src/common/backend/SpotBackendService.test.js
```diff
@@ -155,6 +167,20 @@ describe('SpotBackendService', () => {
                             .toBe(true);
                     });
             });
+            it('notifies the listeners of JWT update - when starting with the registration expired', () => {
+                dateSpy = jest.spyOn(Date, 'now');
+                dateSpy.mockReturnValue(Date.now() + MOCK_RESPONSE.expiresIn);
```
virtuacoplenny: Oh man what is this? I've been using lolex. This looks better.

spot-client/src/common/backend/SpotBackendService.test.js
```diff
@@ -210,6 +236,43 @@ describe('SpotBackendService', () => {
...
+                            mucUrl: 'muc-url',
+                            name: 'mock-muc-name'
+                        }));
+
+                        return spotBackendService.getRoomInfo().then(roomInfo => {
+                            expect(roomInfo).toEqual({
+                                id: 'mock-id',
+                                name: 'mock-muc-name',
+                                roomName: 'muc-url'
+                            });
```
virtuacoplenny: I see a chunk of duplication with the above. Do you think it's worthwhile to move this into a helper?

spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -133,9 +133,14 @@ export class SpotBackendService extends Emitter {
      */
     _maybeClearRegistration(error) {
         if (this.isUnrecoverableRequestError(error)) {
-            this.getJwt() && logger.log('Cleared backend registration');
-            persistence.set(PERSISTENCE_KEY, undefined);
-            this.registration = undefined;
+            if (this.registration) {
+                logger.log('Cleared backend registration');
```
virtuacoplenny: Great move. So obvious this should have been this way.

spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -149,6 +154,10 @@ export class SpotBackendService extends Emitter {
      * @private
      */
     _refreshRegistration(registration) {
+        if (!registration) {
+            throw new Error('No registration object passed to _refreshRegistration');
```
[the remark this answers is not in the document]
paweldomas: right, it should be a rejected promise

spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -225,6 +229,12 @@ export class SpotBackendService extends Emitter {
             + `scheduling refresh to be done in ${delay / ONE_MINUTE} minutes`);
 
         this._refreshTimeout = setTimeout(() => {
+            if (!this.registration) {
+                logger.error('Unable to refresh access token - no registration');
+
+                return;
```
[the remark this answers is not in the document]
paweldomas: If the logic from this commit is ok, then the event would have been emitted whenever this.registration was set to undefined.

spot-client/src/common/backend/SpotBackendService.test.js
paweldomas: right, will do

spot-client/src/common/backend/SpotBackendService.test.js
paweldomas: funny I was trying to find lolex usage, but was not able to find it :P

spot-client/src/common/backend/SpotBackendService.test.js
paweldomas: Okay I gave it a try, please let me know if that's what you had in mind. It's in separate commit and eventually could be left this way or I can squash and rebase if you want me to.

PR #510, opened 2019-08-17
feat(meeting): disconnect temporary remotes on meeting end

virtuacoplenny:


spot-webdriver/specs/adhoc-meeting.spec.js
```diff
@@ -14,7 +14,7 @@ describe('Can start a meeting', () => {
         expect(spotTV.getMeetingName()).toBe(testMeetingName);
     });
 
-    it('and disconnects the remote on meeting end', () => {
+    xit('and disconnects the remote on meeting end', () => {
```
paweldomas: what is the purpose of that? it says "disable while feature is disabled", but it seems to be disabling it permanently, no? How is the feature being enabled figured out?

virtuacoplenny: It's enabled through the config. I could have the test read the config and figure it out from there, if that's what you're implying.

paweldomas: was your plan to enable it later or what? I just dont see a way how this test can be used

virtuacoplenny: Yeah my plan was to un-comment out the test after the feature flag is removed.

paweldomas: this test flag will be false on PR tests anyway, right? so I think it's fine this way

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -653,25 +682,33 @@ export default class XmppConnection {
...
             type = 'join';
         }
 
-        return {
-            from,
-            localUpdate: from === this.getRoomFullJid(),
-            state: Array.from(presence.children)
-                .reduce((acc, child) => {
-                    let value = child.textContent;
+        const state = Array.from(presence.children)
```
paweldomas: Here will be a merge config from https://github.com/jitsi/jitsi-meet-spot/pulls?q=is%3Apr+is%3Aclosed+author%3Apaweldomas

spot-webdriver/specs/adhoc-meeting.spec.js
virtuacoplenny: Yes, the test flag will be false and the test itself is testing the new feature specifically. I could mock out the selector for the flag. Let me look into that.

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -156,6 +161,15 @@ export function connectToSpotTV(joinCode, shareMode) {
 
             history.push(ROUTES.CODE);
 
+            // Assumption alert: kicks can only happen when a temporary remote
+            // has been removed at the end of a meeting.
+            if (error === CONNECTION_EVENTS.KICKED) {
```
paweldomas: Hmm reading this it doesn't sound generic enough. The kick logic is currently XMPP specific, so how about more generic event name? Something around the lines of connection closed by the server or temporary remote disconnected by the server?

spot-client/src/spot-tv/ui/views/meeting.js
```diff
@@ -309,4 +315,36 @@ function mapStateToProps(state) {
...
+        /**
+         * Callback invoked when an error occurs while joining a meeting.
+         *
+         * @param {string} error - An error message to display in an error
+         * notification.
+         * @returns {void}
+         */
+        onError(error) {
+            dispatch(addNotification('error', error));
+
```
paweldomas: Unnecessary new line?

spot-client/src/spot-tv/ui/views/meeting.js
```diff
@@ -221,7 +225,9 @@ export class Meeting extends React.Component {
      */
     _onMeetingLeave(leaveEvent = {}) {
         if (leaveEvent.error) {
-            this.props.dispatch(addNotification('error', leaveEvent.error));
+            this.props.onError(leaveEvent.error);
+        } else {
+            this.props.disconnectAllTemporaryRemotes();
```
paweldomas: How about always trying to disconnect remotes? This operation really should be done by the backend, but since the TV tries to act like one maybe let it always give it a try?

paweldomas: This is not important and I'm not sure about it. I've been wondering if it would cause problem when the network is offline.

spot-client/src/spot-tv/ui/views/meeting.js
virtuacoplenny: Super necessary.

spot-client/src/common/remote-control/xmpp-connection.js
virtuacoplenny: Fixed it but looks like I'll have to send the kicked reasons in some other way, because it's being received as xml.

virtuacoplenny:
Force pushed to resolve conflicts. Did a fixup while I was at it.

spot-client/src/spot-remote/app-state/actions.js
virtuacoplenny: Changed to CLOSED_BY_SERVER. Let me know if it needs more changing.

spot-client/src/spot-tv/ui/views/meeting.js
virtuacoplenny: Do you mean also trying to kick when there is an error joining? I thought it would be annoying to have to re-pair due to an error on join.

If you mean always kicking after a proper hangup, the open source flow does not guarantee the server is an admin, so there could be inconsistent kick behavior. If you're okay with I dont mind attempting the kick.

spot-client/src/common/remote-control/xmpp-connection.js
virtuacoplenny: Instead of changing the way it's sent I changed how it's parsed, using the info you told me about.

spot-client/src/common/remote-control/XmppConnection.test.js
```diff
@@ -58,11 +58,8 @@ describe('XmppConnection', () => {
                 .toEqual({
                     from: IQ_FROM,
                     localUpdate: false,
-                    state: {
-                        isSpot: true,
-                        videoMuted: true
-                    },
-                    type: 'unavailable'
+                    type: 'unavailable',
```
paweldomas: can you add extra special case for unavailable instead of removing a regular one?

spot-client/src/common/remote-control/XmppConnection.test.js
```diff
@@ -94,12 +91,8 @@ describe('XmppConnection', () => {
                 .toEqual({
                     from: IQ_FROM,
                     localUpdate: false,
-                    state: {
-                        isSpot: true,
-                        videoMuted: true,
-                        calendar: calendarArray
-                    },
-                    type: 'unavailable'
+                    type: 'unavailable',
```
paweldomas: same here

spot-client/src/common/remote-control/XmppConnection.test.js
paweldomas: nevermind I got confused by reading commit after commit.

virtuacoplenny:
Force pushed to resolve conflicts in in-call.js caused by adding qaId and share-mode.spec.js caused by changing what the test looks for.

PR #511, opened 2019-08-18
ref(xmpp): remove extra presence listener

virtuacoplenny:


virtuacoplenny:
Works with backend flow selenium tests but fails reconnection tests with open-source flow. I've been poling around general refactoring on the presence handlers with no success, so I'm going to back out of this approach.

PR #512, opened 2019-08-19
ref(xmpp): remove redundant presence listeners

virtuacoplenny:


spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -35,6 +35,7 @@ export default class XmppConnection {
 
         this.initPromise = null;
 
+        this._hasJoinedMuc = true;
```
paweldomas: hmm this doesn't make sense? How can MUC be joined already when the instance is being created?

paweldomas: can you double check the logics? is it possible the things were working without the flag?

virtuacoplenny: You are correct that it should be false. Everything really kind of works without the flag already, but the behavior on master doesn't send out presence on error so I wanted to prevent it.

virtuacoplenny: The previous behavior prevented the situation of a failed MUC join also emitting a disconnect event. The issue would be:
- Remote trying to join a MUC
- Remote fails to join the MUC (initial presence is error)
- The join MUC promise rejects
- The presence callback is also fired for the "error" presence
- "error" presence is considered a disconnect so remote disconnect listeners fire too.

In practice this does not cause that much of a problem right now, although the logs are kind of weird because a disconnect is firing even though the connection was never created.

PR #513, opened 2019-08-19
feature(analytics): log page event when a View is displayed

paweldomas:
Will log a page event when view is being displayed.

Know issue: all Spot-TV setup steps fall under one page as those are modals displayed on the setup view.


spot-client/src/common/ui/views/view.js
```diff
@@ -27,6 +28,7 @@ class View extends React.Component {
      */
     componentDidMount() {
         logger.log('View mounted', { name: this.props.name });
+        analytics.page(this.props.name);
```
virtuacoplenny: I thought you wanted analytics to be done in middleware.

PR #515, opened 2019-08-20
Fix waiting for spot tv

paweldomas:


spot-webdriver/user/spot-tv-user.js
```diff
@@ -55,6 +55,15 @@ class SpotTV extends SpotUser {
         return meetingPage.getMeetingName();
     }
 
+    /**
+     * Retrieves the short lived pairing code displayed on the calendar page.
+     *
+     * @returns {string}
+     */
+    getShortLivedPairingCode() {
+        return this.getCalendarPage().getJoinCode();
```
virtuacoplenny: I wonder if this method is necessary. I think the other tests use the SpotUser to get other page objects, and don't use pass-through methods, with `getMeetingName` being the exception that maybe needs fixing.

paweldomas: It seems necessary to me when I'm too lazy to do those extra calls every time join code needs to be fetched. Also if things change it will be just this one place to modify, as it removes duplication(even though a small one).

paweldomas: I also see this one different to the other page objects usages you mention. The other usages are actually interested in exercising specific parts of the UI. In this context I'm interested in getting the join code and I don't care where it is. It could be even pre-configured.

PR #516, opened 2019-08-20
fix(ci): re-enable webdriver tests

virtuacoplenny:


scripts/ci.sh
```diff
@@ -1,39 +1,68 @@
...
+    # eating into its own run time.
+    echo "Sleeping to give webpack-dev-server time to start"
+    sleep 30
+    echo "Completed giving webpack-dev-server time to start"
 
-    # python -m SimpleHTTPServer $TEST_PORT &
-    # pid=$!
+    echo "Starting webdriver"
+    cd ../spot-webdriver
+    npm start || webdriver_exit_code=$?
```
paweldomas: how does this line work:
`npm start || webdriver_exit_code=$?`
?

paweldomas:
The script looks good to me other than the fact that the tests are failing, but I guess you're aware of that.

virtuacoplenny:
The test failures are all related to screensharing. I've added more test reporters to record failures. I'll keep debugging.

virtuacoplenny:
The test failures seem to be an issue where the tests try to screensharing but there is no support for screensharing. This is possibly and http vs https issue. I'm enabling https for the webpack dev server. If this fixes it, I'll make separate commits to address https and some other issues I've encountered.

virtuacoplenny:
jenkins is running an older version of chrome (v. 69) so the tests are failing.

virtuacoplenny:
Should we keep the "https" change? I'm okay with it but I'll have to make a separate commit out of principle. There will be an invalid cert error though.

virtuacoplenny:
Jenkins test this please.

paweldomas:
If it works without the https then lets not keep it. The invalid cert warning is kind of annoying when testing. Does it actually apply to the `npm run start:dev` as well or not? If not then I dont care.

virtuacoplenny:
Okay, I think it's ready now. I have jenkins configured to use the junit results and store the timelime as an archive, and I installed a workspace cleanup plugin so the workspace gets cleared with each run (which will likely increase build times a tick or several).

PR #523, opened 2019-08-23
ref(ultrasound): remove workaround of playing on interaction

virtuacoplenny:
Before it was known a mobile application was required,
ultrasound was built to be played from mobile safari,
which requires a user interaction before playing sound.
The workaround was to start playing sound on join code
submit, but that is not needed anymore as the react
native webview can allow sound to play without
interaction.

Probably a larger refactoring of ultrasound is needed to
move away from the "loader" pattern.

### DIFF
--- a/spot-client/src/spot-remote/ui/views/join-code-entry.js
+++ b/spot-client/src/spot-remote/ui/views/join-code-entry.js
@@ -22,7 +22,6 @@ import {
     setAPiReceivedJoinCode
 } from './../../app-state';
 import { NavButton, NavContainer } from './../components';
-import { withUltrasound } from './../loaders';
 
 /**
  * Displays a view to enter a join code for connecting with a Spot-TV instance.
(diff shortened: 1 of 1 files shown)


spot-client/src/spot-remote/ui/views/join-code-entry.js
```diff
@@ -78,7 +76,6 @@ export class JoinCodeEntry extends React.Component {
      */
     componentDidMount() {
         this.props.updateReadyStatus(true);
-        this.props.ultrasoundService.setMessage('');
```
paweldomas: is that not needed anymore, because ultrasound listens for Redux state and it will track the join code changes?

virtuacoplenny: In current master, withUltrasound already sets the message to undefined on unmount. To when the remote control view unmounts it should already be clearing the message. The way I'm interpreting the code right now is this call to set an empty string is redundant.

PR #530, opened 2019-08-28
feat(feedback): show feedback button outside of meeting

virtuacoplenny:
feat(feedback): show feedback button outside of meeting

- Wrap all Spot-Remote route displaying to show the
  global feedback button.
- Create a global feedback button for Spot-Remote and
  Spot-TV (setup only) to use.
- Move feedback analytics to shared middleware as
  now both TV and Remote can send feedback.
- New redux state for controlling app feedback display.
  Call feedback display has not been changed.
- Move Feedback components into common and modify
   so the star rating can be skipped automatically.

Screen Shot 2019-08-28 at 3 10 04 PM
Screen Shot 2019-08-28 at 3 01 00 PM
Screen Shot 2019-08-28 at 3 00 47 PM

### DIFF
--- a/spot-client/src/app.js
+++ b/spot-client/src/app.js
@@ -9,6 +9,8 @@ import { ROUTES } from 'common/routing';
 import {
     ErrorBoundary,
     FatalError,
+    FeedbackOpener,
+    FeedbackOverlay,
     IdleCursorDetector,
     ModalManager,
     Notifications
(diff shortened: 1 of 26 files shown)


paweldomas:
It would be nice that feedback dialog would have the background, but not a big deal

virtuacoplenny:
Noticed the ios keyboard wasn't properly pushing the textarea up and would end up covering it. Workaround for now, so this can be deployed, is leaving the text area higher.
Screen Shot 2019-08-28 at 5 44 09 PM

PR #531, opened 2019-08-29
Test fixes

paweldomas:


spot-webdriver/user/spot-user.js
```diff
@@ -21,6 +21,44 @@ class SpotUser {
...
+    /**
+     * Cleanups the session.
+     *
+     * @returns {void}
+     */
+    cleanup() {
+        this.setNetworkOnline();
+        this.disconnectRemoteControlService();
+        this.clearStorage();
+        this.stop();
```
virtuacoplenny: No problems were encountered with stop? I had tried doing the same logic as stop before (but also before the manual disconnected of RCS) and the page would get hung.

paweldomas: Hmm no, I haven't seen that at least not recently. I've seen that some time ago when closing page after disconnect was done synchronous.

PR #532, opened 2019-08-29
Misc fixes

paweldomas:


spot-client/src/common/utils/retry.js
```diff
@@ -10,6 +10,6 @@
  * @returns {number} - The amount of waiting before trying another time given in milliseconds.
  * @private
  */
-export function getJitterDelay(retry, minDelay = 500, base = 1) {
+export function getJitterDelay(retry, minDelay = 500, base = 2) {
```
virtuacoplenny: So obvious once you point it out.

PR #533, opened 2019-08-29
No reconnect in RCS

paweldomas:


spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -227,10 +228,6 @@ export function createSpotTVRemoteControlConnection({ pairingCode, retry }) {
             SERVICE_UPDATES.REMOTE_JOIN_CODE_CHANGE,
             onRemoteJoinCodeChange
         );
-        remoteControlServer.addListener(
-            SERVICE_UPDATES.RECONNECT_UPDATE,
```
virtuacoplenny: Probably can remove the constant.

spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -156,25 +161,21 @@ export function createSpotTVRemoteControlConnection({ pairingCode, retry }) {
...
 
                 throw error;
-            } else if (retry) {
-                doConnect();
+            } else if (retry || initiallyConnected) {
+                const jitter = getJitterDelay(3);
+
+                logger.log(`Spot TV will try to reconnect after ${jitter}ms`);
+
+                return new Promise((resolve, reject) => {
```
virtuacoplenny: What is the promise for?

spot-client/src/spot-remote/ui/views/remote-control.js
```diff
@@ -52,7 +51,6 @@ export class RemoteControl extends React.PureComponent {
             <View name = 'remoteControl'>
                 { this._getView() }
                 <ElectronDesktopPickerModal />
-                <ReconnectOverlay />
```
virtuacoplenny: With the overlay gone and the loading screen displaying, does the in-meeting reconnect work okay? I can see a time where maybe a prosody deploy has to happen and connection lost momentarily while in meeting.

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -141,8 +152,6 @@ export function connectToSpotTV(joinCode, shareMode) {
 
             _clearSubscriptions();
 
-            dispatch(clearSpotTVState());
```
virtuacoplenny: Why not clear spot tv state?

spot-client/src/common/app-state/remote-control-service/reducer.js
```diff
@@ -43,12 +48,6 @@ const remoteControlService = (state = DEFAULT_STATE, action) => {
             joinWithScreensharing: action.screensharingType
         };
 
-    case RECONNECTING:
-        return {
-            ...state,
-            isReconnecting: action.isReconnecting
```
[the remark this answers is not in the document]
paweldomas: thanks! I have missed that one

spot-client/src/spot-remote/ui/views/remote-control.js
paweldomas: I will check that

spot-client/src/spot-remote/app-state/actions.js
paweldomas: hmm I think I was fighting some issue and have it fixed in another way and this stayed. Will try to undo the change.

spot-client/src/spot-tv/app-state/actions.js
paweldomas: I was trying to keep the method's signature. I see it is used as `catch` in `doConnect` at line 220. I'm not sure about that. Do you think it's safe to remove the promise?

spot-client/src/common/app-state/remote-control-service/reducer.js
```diff
@@ -29,6 +29,11 @@ const remoteControlService = (state = DEFAULT_STATE, action) => {
 
     case CREATE_CONNECTION:
         return updateStateForAsyncAction(state, 'connect', action);
+    case DESTROY_CONNECTION:
+        return {
+            ...state,
+            connect: undefined
```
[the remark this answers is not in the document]
paweldomas: 'connect' is the state created by the asynchronous action 'CREATE_CONNECTION'. It needs to be cleared in order for the method `isConnectionEstablished`(`common/app-state/remote-control-service`) to stop return `true` after connection is closed. Do you have ideas on how can this be improved? I can also put that explanation in the comment.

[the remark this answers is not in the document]
virtuacoplenny: Oh I see. Comment for now seems good. I was looking at the default props but now I realize I just screwed up those default props by not adding more.

spot-client/src/spot-tv/app-state/actions.js
virtuacoplenny: I see. You actually read docs like a good person. And the throws are how it "returns" a rejected promise. Looks good as is.

spot-client/src/spot-remote/ui/views/remote-control.js
virtuacoplenny: We could also redirect back home maybe if that's easier somehow.

paweldomas: Okay glad you've asked me to check on that, because I've found serious bug with a reconnect loop(see the last squash commit)

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -206,9 +207,13 @@ export class BaseRemoteControlService extends Emitter {
...
-        this.disconnect()
-            .then(() => this.emit(
-                SERVICE_UPDATES.UNRECOVERABLE_DISCONNECT, reason));
+        // If XMPP server is not working calling disconnect triggers onDisconnect again.
+        if (!this._disconnecting) {
+            this._disconnecting = true;
+            this.disconnect()
+                .then(() => this.emit(
+                    SERVICE_UPDATES.UNRECOVERABLE_DISCONNECT, reason));
+        }
```
virtuacoplenny: Should disconnecting be set to false after disconnecting is finished?

paweldomas: I don't think so. Once disconnect has been called it is not safe to call it again until the service is connected again.

paweldomas:
Something's weird with the tests I'll be looking into that.

spot-client/src/spot-remote/ui/views/remote-control.js
virtuacoplenny: What happens to the TV when in-meeting and the disconnect occurs and then the connection recovers?

paweldomas: So I was testing the case where RCS prosody is restarted to simulate disconnect on the remote control plane. Both TV and remote(a permanent one) will come back to the meeting and it looks good.

Now I will try another case where will just turn to "offline" in the network conditioner so that both RCS prosody and the meeting prosody will disconnect and see what happens.

paweldomas: Yes, it's all good on the in-meeting reconnects. TV comes back to the meeting and remote reconnect correctly to the Spot TV.

PR #534, opened 2019-08-30
feat(dtmf): add dial pad in call

virtuacoplenny:
- New RCS command for playing touch tones
- New DTMF button for opening the DTMF Modal, which shows a dial pad
~- New StatelessDialPad, which is a copy of DialPad, which does not store internal state of the entered phone number. Used so DTMFModal can clear the entered phone number after submit.~
~- Ended up moving screensharing start functionality into the screensharing button as a result of temporary period where the screensharing button was supposed to be in the More Modal.~

~I'll try to see about splitting out some of these commits. Regardless, these changes cannot be merged until DTMF is fixed on jitsi-meet and the changes have been deployed to all necessary environments.~ Done.

### DIFF
--- a/spot-client/src/common/app-state/config/default-config.js
+++ b/spot-client/src/common/app-state/config/default-config.js
@@ -228,7 +228,8 @@ export default {
     TEMPORARY_FEATURE_FLAGS: {
         /**
          * Allow DTMF support to be hidden while waiting for the jitsi-meet side
-         * to be deployed and dial tone playing fixed.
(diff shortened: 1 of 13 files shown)


virtuacoplenny:
Okay, the refactorings are merged in. I'll see about hiding the DTMF button behind a temporary config flag.

virtuacoplenny:
Screen Shot 2019-09-04 at 9 22 52 AM

PR #536, opened 2019-08-30
ref(dial-pad): add stateless version

virtuacoplenny:


virtuacoplenny:
Technically they're both used. The DialPad is used on the waiting screen by the remote. The StatelessDialPad is used by the DialPad. In the future I want the DTMF modal to use the StatelessDialPad.

PR #538, opened 2019-09-03
feat: display some debugging steps when waiting for spot-tv

virtuacoplenny:
- Use the StatusOverlay styling to make message more visible
- Provide an explicit instructions to check if the Spot-TV
  is on
- Adjust z-indexes so feedback icon displays above the new
  status overlay

Screen Shot 2019-09-03 at 10 56 56 AM


virtuacoplenny:
Screen Shot 2019-09-03 at 1 22 03 PM

PR #545, opened 2019-09-06
ref(XmppConnection): send Spot TV presence as single JSON string

paweldomas:
Will stringify the entire Spot TV status object into single JSON string and put it in a custom XML element inside MUC presence. The status will come out of XMPPConnection as a ready to use JS object.

### DIFF
--- a/spot-client/src/common/remote-control/xmpp-connection.js
+++ b/spot-client/src/common/remote-control/xmpp-connection.js
@@ -6,6 +6,18 @@ import { JitsiMeetJSProvider } from 'common/vendor';
 import { IQ_NAMESPACES, IQ_TIMEOUT } from './constants';
 
 /**
+ * XML element name for spot status added to MUC presence.
+ * @type {string}
+ */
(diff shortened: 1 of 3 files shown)


spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -657,21 +663,24 @@ export default class XmppConnection {
...
+            state = statusElement ? JSON.parse(statusElement.textContent) : { };
+        } catch (error) {
+            logger.error('Failed to parse presence', {
+                from,
+                type
+            });
 
-                    acc[child.tagName] = value;
+            return undefined;
+        }
```
virtuacoplenny: The weird part will be if there is a time when people try to connect a new remote control to a TV running the old code. I wonder if we need to include some transitionary code to account for that edge case.

paweldomas: yeah.. good point

PR #552, opened 2019-09-10
feat(dial): press 0 for +

virtuacoplenny:


spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -40,7 +40,7 @@ export default class StatelessDialPad extends React.Component {
         this._onDialButtonClick = this._onDialButtonClick.bind(this);
         this._onGoToCall = this._onGoToCall.bind(this);
         this._onInputChange = this._onInputChange.bind(this);
-        this._renderDialButton = this._renderDialButton.bind(this);
+        this._onReplaceLastChar = this._onReplaceLastChar.bind(this);
```
paweldomas: why `_renderDialButton` bind was removed? it's used hasn't changed

paweldomas: was it not necessary? Im too lazy to read the code

virtuacoplenny: Correct, the binding is not necessary. I didn't realize I left this change in though and it'll be better if I revert the change to keep the PR a bit more pure.

virtuacoplenny:
Screen Shot 2019-09-11 at 9 43 29 AM

PR #560, opened 2019-09-16
feat(dial pad): pretty format and validate phone numbers

paweldomas:
Use libphonenumber-js to format the displayed phone number as it's being
typed. Also validates if the phone number is valid, before calling
the onSubmit callback.

Uses 'US' as a default country code for formatting and validation if no
country code is entered manually.

Modifies StatelessDialPad to not pass the value and only call onSubmit
callback when submit action is executed. It is not supposed to track
the state anyway and it is confusing whether to use the value given on
submit or the value that's been submitted last in the on change
callback.

### DIFF
--- a/spot-client/src/spot-remote/ui/components/dial-pad/DialPad.test.js
+++ b/spot-client/src/spot-remote/ui/components/dial-pad/DialPad.test.js
@@ -16,47 +16,103 @@ describe('DialPad', () => {
     });
 
     test('entering a number with a mouse', () => {
-        for (let i = 1; i < 4; i++) {
-            dialPad.find(`#dial-button-${i}`).simulate('mousedown');
-        }
+        dialPad.find(`#dial-button-${2}`).simulate('mousedown');
+        dialPad.find(`#dial-button-${2}`).simulate('mousedown');
+        dialPad.find(`#dial-button-${2}`).simulate('mousedown');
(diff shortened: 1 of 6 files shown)


spot-client/src/spot-remote/ui/components/dial-pad/DialPad.test.js
```diff
@@ -16,47 +16,103 @@ describe('DialPad', () => {
...
+        dialPad.find(`#dial-button-${4}`).simulate('touchstart');
+
+        dialPad.find(`#dial-button-${3}`).simulate('touchstart');
+        dialPad.find(`#dial-button-${3}`).simulate('touchstart');
+        dialPad.find(`#dial-button-${3}`).simulate('touchstart');
+
+        dialPad.find(`#dial-button-${4}`).simulate('touchstart');
+        dialPad.find(`#dial-button-${4}`).simulate('touchstart');
+        dialPad.find(`#dial-button-${4}`).simulate('touchstart');
+        dialPad.find(`#dial-button-${4}`).simulate('touchstart');
```
virtuacoplenny: Maybe instead have a function to loop over a string?

paweldomas: yeah

PR #563, opened 2019-09-16
feat: integrate 'countryCode' provided by the backend with dial pad

paweldomas:
Will use ISO 3166-1 alpha-2 country code provided by the backend for phone number formatting in the dial pad.


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -54,12 +57,46 @@ export default class DialPad extends React.Component {
...
 
+    /**
+     * Initializes the {@code AsYouType} utility for phone number fomratting with the given country code.
+     *
+     * @param {string} countryCode - ISO 3166-1 alpha-2 country code.
+     * @private
+     * @returns {void}
+     */
+    _createAsYouType(countryCode) {
+        logger.log(`Init dial pad with country code: ${countryCode}`, { countryCode });
```
virtuacoplenny: nit: should not include countryCode in the log string as it's already reported in the context object.

spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -54,12 +57,46 @@ export default class DialPad extends React.Component {
             formattedPhone: ''
         };
 
-        this._asYouType = new AsYouType('US');
+        this._createAsYouType(props.countryCode);
```
virtuacoplenny: nit: maybe add a default prop of US? It's more reacty to declare the default, I think.

spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
paweldomas: Actually I did this intentional, because Im lazy to unwrap logs when reading them. I did this in few other places recently, so if you are really against that I will stop doing it.

spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
paweldomas: ok

spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
virtuacoplenny: We were supposed to use a constant string for the log message for some querying purpose; that's what I was told anyway. Let's just go with what's easier for now.

PR #574, opened 2019-09-18
feat(dial pad): dynamically enable/disable the call button

paweldomas:
The dial pad will enabled the call button only when a valid phone number has been entered.

### DIFF
--- a/spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
+++ b/spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
@@ -49,11 +49,6 @@ export class DialPad extends React.Component {
         super(props);
 
         this.state = {
-            /**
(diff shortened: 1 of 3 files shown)


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -224,9 +242,10 @@ export class DialPad extends React.Component {
         }
 
         this.setState({
+            formattedPhone,
+            isPhoneNumberValid: typeof this._getPhoneNumber() === 'string',
```
virtuacoplenny: Does this need to be stored in state?

paweldomas: no

PR #576, opened 2019-09-18
Dial out indication

paweldomas:
Screenshot 2019-09-18 16 09 28


spot-client/src/common/app-state/spot-tv/actions.js
```diff
@@ -41,3 +41,22 @@ export function setSpotTVState(newState) {
...
+ * @param {Array<{ type: string, number: string }>} [invites] - An array of invites as specified by Jitsi Meet iframe
+ * API.
+ * @returns {Function}
+ */
+export function storePhoneNumberFromInvites(invites) {
+    return dispatch => {
+        const phoneInvites = invites && invites.filter(invite => invite.type === 'phone');
+        const phone = phoneInvites && phoneInvites[0] && phoneInvites[0].number;
+
+        dispatch(setSpotTVState({ invitedPhoneNumber: phone }));
```
virtuacoplenny: I don't think this needs to be wrapped in an extra function because I see no async that warrants it.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
```diff
@@ -111,11 +114,16 @@ export class InCall extends React.Component {
         }
 
         const { meetingName } = parseMeetingUrl(inMeeting);
+        const phoneNumber = this.props.invitedPhoneNumber && parsePhoneNumberFromString(this.props.invitedPhoneNumber);
```
virtuacoplenny: Maybe this should be passed in already parsed, if easy? Or maybe a separate component that does the parsin?

spot-client/src/spot-tv/ui/views/home.js
```diff
@@ -174,7 +174,7 @@ export class Home extends React.Component {
             let path = `${ROUTES.MEETING}?location=${data.meetingName}`;
 
             if (data.invites) {
-                path += `&invites=${JSON.stringify(data.invites)}`;
+                path += `&invites=${encodeURIComponent(JSON.stringify(data.invites))}`;
```
virtuacoplenny: nice catch

spot-client/src/common/app-state/spot-tv/actions.js
```diff
@@ -53,8 +53,7 @@ export function setSpotTVState(newState) {
  */
 export function storePhoneNumberFromInvites(invites) {
     return dispatch => {
-        const phoneInvites = invites && invites.filter(invite => invite.type === 'phone');
-        const phone = phoneInvites && phoneInvites[0] && phoneInvites[0].number;
+        const phone = invites?.filter(invite => invite.type === 'phone')?.[0]?.number;
```
virtuacoplenny: This could be done as a fine instead for now.

spot-client/src/common/app-state/spot-tv/actions.js
paweldomas: yeah, I was stupid thinking there's no other way for it to work 🤦‍♂ but actually it should work without the async just fine

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
paweldomas: hmm this parsing is really UI formatting so I would rather pass the data as raw as possible and do the formatting in the UI component

spot-client/src/common/app-state/spot-tv/actions.js
paweldomas: oh okay

spot-client/src/common/app-state/spot-tv/actions.js
```diff
@@ -49,13 +49,11 @@ export function setSpotTVState(newState) {
  *
  * @param {Array<{ type: string, number: string }>} [invites] - An array of invites as specified by Jitsi Meet iframe
  * API.
- * @returns {Function}
+ * @returns {Object}
  */
 export function storePhoneNumberFromInvites(invites) {
-    return dispatch => {
-        const phone = invites?.filter(invite => invite.type === 'phone')?.[0]?.number;
+    const phone = invites?.find(invite => invite.type === 'phone')?.number;
```
virtuacoplenny: style nit: maybe a default arg for invites? That way it's always defined.

spot-client/src/spot-remote/ui/utils/formatPhoneNumber.js
```diff
@@ -0,0 +1,11 @@
+import { parsePhoneNumberFromString } from 'libphonenumber-js/max';
+
+/**
+ * Pretty prints given phone number. Will return a formatted text if a valid phone number string is given.
+ *
+ * @param {string} [phoneNumber] - A not pretty phone number string, for example: '+12343334444'.
+ * @returns {?string}
+ */
+export function formatPhoneNumber(phoneNumber) {
+    return (phoneNumber && parsePhoneNumberFromString(phoneNumber))?.formatInternational();
```
virtuacoplenny: What would happen if undefined or an empty string is passed into parsePhoneNumberFromString? That could remove some of the defensive checking.

spot-client/src/common/app-state/spot-tv/actions.js
```diff
@@ -41,3 +41,19 @@ export function setSpotTVState(newState) {
...
+ * Spot TV state so that Spot Remotes can display it in the UI.
+ *
+ * @param {Array<{ type: string, number: string }>} [invites] - An array of invites as specified by Jitsi Meet iframe
+ * API.
+ * @returns {Object}
+ */
+export function storePhoneNumberFromInvites(invites) {
+    const phone = invites?.find(invite => invite.type === 'phone')?.number;
+
+    return setSpotTVState({ invitedPhoneNumber: phone });
```
virtuacoplenny: You want this to be set to undefined if no invite is found, correct?

paweldomas:
I will probably rebase and force push because those are separate commits

spot-client/src/common/app-state/spot-tv/actions.js
paweldomas: yes

spot-client/src/common/app-state/spot-tv/actions.js
paweldomas: yeah, but too lazy to change it given `.?` is now so easy

spot-client/src/spot-remote/ui/utils/formatPhoneNumber.js
paweldomas: it throws that's why it needs to be there

PR #578, opened 2019-09-19
Fix use of base URL

saghul:


spot-client/src/spot-remote/ui/components/screenshare/ScreensharePicker.js
```diff
@@ -294,7 +294,7 @@ export class ScreensharePicker extends React.Component {
                         To share, use Chrome desktop and go to
                     </div>
                     <div className = 'share-url'>
-                        { `${shareDomain || windowHandler.getHost()}/` }
+                        { `${shareDomain || windowHandler.getBaseUrl()}/` }
```
virtuacoplenny: Is it tricky to get ScreensharePicker to use join-info? Maybe the styling isn't set up for it.

spot-client/src/spot-tv/ui/components/join-info/join-info.js
```diff
@@ -56,21 +56,11 @@ class JoinInfo extends React.Component {
 
         return (
             <>
-                { `${shareDomain || windowHandler.getHost()}/` }
+                { `${shareDomain || windowHandler.getBaseUrl()}/` }
```
virtuacoplenny: Probably at some point the same change will have to be made for shareDomain somehow, but not today.

spot-client/src/spot-remote/ui/components/screenshare/ScreensharePicker.js
saghul: They do behave a bit different. In one case it's possible to hide the domain, not so on the other. In addition, one set qaId (I don't know what that is) and the other one does not.

I'm kinda on the clock, but I pink-promise you I'll take care of it while at the conference, if w find a way forward.

virtuacoplenny: qaid is an attribute set on elements to be used for the selenium tests to look up instead of depending on dom or other attributes like id or class.

virtuacoplenny: There's also a redesign in flight so probably I'd take care of it at some point anyways, if I remember there's duplication.

PR #579, opened 2019-09-19
feat(dial pad): check with the phone authorize service before dialing

paweldomas:
Will check entered phone number with the phone authorize service before calling a number that could have been blacklisted.

Screenshot 2019-09-19 16 13 32(2)
Screenshot 2019-09-19 16 13 38(2)

### DIFF
--- a/spot-client/src/common/backend/utils.js
+++ b/spot-client/src/common/backend/utils.js
@@ -234,15 +234,15 @@ export function getRemotePairingCode(serviceEndpointUrl, jwt) {
  *
  * @param {string} serviceEndpointUrl - The URL pointing to the phone authorize service.
  * @param {string} phoneNumber - E.164 formatted phone number to be checked.
- * @returns {Promise<{allow: boolean, message: string}>}
+ * @returns {Promise<void>}
  */
 export function phoneAuthorize(serviceEndpointUrl, phoneNumber) {
     if (!serviceEndpointUrl) {
-        throw new Error('phoneAuthorize - serviceEndpointUrl is required');
+        return Promise.reject('phoneAuthorize - serviceEndpointUrl is required');
     }
 
     if (!phoneNumber) {
-        throw new Error('phoneAuthorize - phoneNumber is required');
+        return Promise.reject('phoneAuthorize - phoneNumber is required');
     }
 
     const requestOptions = {
--- a/spot-client/src/common/backend/utils.js
+++ b/spot-client/src/common/backend/utils.js
@@ -230,6 +230,44 @@ export function getRemotePairingCode(serviceEndpointUrl, jwt) {
 }
 
 /**
+ * Checks if it's ok to call given phone number.
+ *
+ * @param {string} serviceEndpointUrl - The URL pointing to the phone authorize service.
+ * @param {string} phoneNumber - E.164 formatted phone number to be checked.
+ * @returns {Promise<{allow: boolean, message: string}>}
+ */
+export function phoneAuthorize(serviceEndpointUrl, phoneNumber) {
+    if (!serviceEndpointUrl) {
+        throw new Error('phoneAuthorize - serviceEndpointUrl is required');
+    }
+
+    if (!phoneNumber) {
+        throw new Error('phoneAuthorize - phoneNumber is required');
+    }
+
+    const requestOptions = {
+        headers: createHeaders(),
+        method: 'GET',
+        mode: 'cors'
+    };
+
+    return fetchWithRetry({
+        operationName: 'phoneAuthorize',
+        requestOptions,
+        url: `${serviceEndpointUrl}?phone=${encodeURIComponent(phoneNumber)}`
+    })
+        .then(({ message, allow }) => {
+            if (typeof allow === 'undefined') {
+                throw new Error('phoneAuthorize - no "allow" field in the response');
+            }
+
+            if (!allow) {
+                throw message;
+            }
+        });
+}
+
+/**
  * @typedef {Object} RefreshTokenResponse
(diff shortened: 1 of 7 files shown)


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -244,6 +262,28 @@ export class DialPad extends React.Component {
...
+function mapDispatchToProps(dispatch) {
+    return {
+        /**
+         * Display an app notification.
+         *
+         * @param {string} type - The type of the notification to display.
+         * @param {string} message - The text to display in the notification.
+         * @returns {void}
+         */
+        onAddNotification(type, message) {
```
virtuacoplenny: I think conceptually this prop should be called something like onPhoneValidationError, and not mention something about a notification because that's an implementation detail that doesn't need to be stated in the prop name.

spot-client/src/common/backend/utils.js
```diff
@@ -229,6 +229,44 @@ export function getRemotePairingCode(serviceEndpointUrl, jwt) {
...
+/**
+ * Checks if it's ok to call given phone number.
+ *
+ * @param {string} serviceEndpointUrl - The URL pointing to the phone authorize service.
+ * @param {string} phoneNumber - E.164 formatted phone number to be checked.
+ * @returns {Promise<{allow: boolean, message: string}>}
+ */
+export function phoneAuthorize(serviceEndpointUrl, phoneNumber) {
+    if (!serviceEndpointUrl) {
+        throw new Error('phoneAuthorize - serviceEndpointUrl is required');
```
virtuacoplenny: The function above returns Promise.reject and the function below throws errors. Which one would you prefer? Is there a difference?

spot-client/src/common/backend/utils.js
```diff
@@ -229,6 +229,44 @@ export function getRemotePairingCode(serviceEndpointUrl, jwt) {
     });
 }
 
+/**
+ * Checks if it's ok to call given phone number.
+ *
+ * @param {string} serviceEndpointUrl - The URL pointing to the phone authorize service.
+ * @param {string} phoneNumber - E.164 formatted phone number to be checked.
+ * @returns {Promise<{allow: boolean, message: string}>}
```
virtuacoplenny: I think this jsdoc is incorrect. I think the promise returns nothing on resolve.

paweldomas: oh yes it got outdated at some point when simplifying

spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
paweldomas: ok

spot-client/src/common/backend/utils.js
paweldomas: yeah, it needs to be rejected promise

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -136,10 +137,12 @@ export default class StatelessDialPad extends React.Component {
                         <div className = 'dial-pad-footer-button'>
                             <button
                                 className = { callButtonClassName }
-                                disabled = { this.props.disableCallButton }
+                                disabled = { this.props.disableCallButton || this.props.dialingInProgress }
```
virtuacoplenny: Shouldn't disableCallButton be set to true instead of also relying on dialingInProgress? Seems like a redundant prop.

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -136,10 +137,12 @@ export default class StatelessDialPad extends React.Component {
...
                                 onClick = { this._onSubmit }
                                 type = 'submit'>
-                                <Call />
+                                { this.props.dialingInProgress
+                                    ? <div className = 'dialing-dots' >...</div>
```
virtuacoplenny: Is it not possible to use the loading-icon component? Does it not look okay?

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
paweldomas: There are two states "disabled call button" and "loading icon on call button".

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
paweldomas: I will try that

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
paweldomas: maybe I can add extra variable on the top which will control the button's disabled state? would that make things any better?

paweldomas:
Screenshot 2019-09-20 10 47 54

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -7,6 +7,7 @@ import { Backspace, Call, KeyboardArrowDown } from 'common/icons';
 import CountryCodePicker from './CountryCodePicker';
 import DialButton from './dial-button';
 import NumberInput from './NumberInput';
+import LoadingIcon from '../../../../common/ui/components/loading-icon/loading-icon';
```
paweldomas: oh wait maybe I can make this import not direct

PR #587, opened 2019-09-23
fix(Spot Remote): waiting for TV to connect appear to often

paweldomas:
Moves the before unload handler to index where 'destroyConnection' action can be dispatched before starting the disconnect. This allows to combine 'isConnectionEstablished' selector for the waiting for spot TV message condition(because it clears the connection established state).

### DIFF
--- a/spot-client/src/index.js
+++ b/spot-client/src/index.js
@@ -33,7 +33,10 @@ import {
     loadScript,
     setPersistedState
 } from 'common/utils';
+
+import { disconnectFromSpotTV } from 'spot-remote/app-state';
 import { ExternalApiSubscriber } from 'spot-remote/external-api';
+import { disconnectSpotTvRemoteControl } from './spot-tv/app-state';
 
 import App from './app';
 
@@ -97,6 +100,16 @@ const analyticsAppKey = getAnalyticsAppKey(reduxState);
 if (analyticsAppKey) {
     analytics.addHandler(new SegmentHandler(deviceId, analyticsAppKey));
 }
+
+window.addEventListener(
+    'beforeunload',
+    event => {
+        store.dispatch(disconnectSpotTvRemoteControl(event)).catch(error => {
+            console.error('Failed to disconnect Spot TV on unload', error);
+        });
+        store.dispatch(disconnectFromSpotTV(event));
+    }
+);
 
 // eslint-disable-next-line max-params
 window.onerror = (message, url, lineNo, columnNo, error) => {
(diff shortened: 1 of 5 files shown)


spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -280,13 +280,18 @@ function _onSpotTVStateChange({ dispatch }, data) {
...
-export function disconnectFromSpotTV() {
+export function disconnectFromSpotTV(event) {
     return dispatch => {
         _clearSubscriptions();
 
-        remoteControlClient.disconnect();
+        dispatch(destroyConnection());
+
+        remoteControlClient.disconnect(event).catch(error => {
+            logger.error('Failed to disconnect Spot Remote', { error });
```
virtuacoplenny: When does this error happen?

spot-client/src/spot-remote/ui/views/remote-control.js
```diff
@@ -68,7 +70,7 @@ export class RemoteControl extends React.PureComponent {
      * @returns {ReactElement}
      */
     _getView() {
-        if (!this.props.isConnectedToSpot) {
+        if (this.props.isConnectionEstablished && !this.props.isConnectedToSpot) {
```
virtuacoplenny: Could isConnectionEstablished and isConnectedToSpot be combined into one prop instead? Maybe just a "loading" or "isConnecting" prop.

spot-client/src/index.js
```diff
@@ -98,6 +101,16 @@ if (analyticsAppKey) {
     analytics.addHandler(new SegmentHandler(deviceId, analyticsAppKey));
 }
 
+window.addEventListener(
+    'beforeunload',
+    event => {
+        store.dispatch(disconnectSpotTvRemoteControl(event)).catch(error => {
+            console.error('Failed to disconnect Spot TV on unload', error);
```
virtuacoplenny: Why is this catch for `disconnectSpotTvRemoteControl` in index.js but the catch for `disconnectFromSpotTV` is in actions.js?

paweldomas: because one returns a promise and the other one does not, but maybe I can clean this up

spot-client/src/spot-remote/app-state/actions.js
paweldomas: I don't know, but I was trying to preserve the behavior of the handler that's being removed from the RCS in this commit.

PR #594, opened 2019-09-26
Dtmf dial pad changes

paweldomas:


spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -126,7 +127,7 @@ export default class StatelessDialPad extends React.Component {
                         { this._renderDialButton('9', 'WXYZ') }
                     </div>
                     <div className = 'row'>
-                        { this._renderDialButton('＊') }
+                        { this._renderDialButton('*') }
```
virtuacoplenny: Does this look okay? I tried matching the designs and the asterisk didn't look good, which is what led to the bug. I was thinking we might need to add support for showing one value but having another value that's actually used.

paweldomas: Okay I can try doing that

paweldomas: Yeah that character looks better somehow

spot-client/src/spot-remote/ui/components/dial-pad/StatelessDialPad.js
```diff
@@ -185,7 +186,11 @@ export default class StatelessDialPad extends React.Component {
      * @returns {void}
      */
     _onDialButtonClick(value) {
-        this.props.onChange(`${this.props.value}${value}`);
+        // The ＊ sign was used as the buttons value to make it look better. But it needs to be translated back to * in
+        // order for the DTMF tones to work correctly.
+        const filteredValue = value === '＊' ? '*' : value;
+
+        this.props.onChange(`${this.props.value}${filteredValue}`);
```
virtuacoplenny: I was thinking of letting the button component handle this but this works too.

PR #604, opened 2019-10-02
fix(calendar): refire fetch event on success after error

virtuacoplenny:


spot-client/src/spot-tv/calendars/calendarService.js
```diff
@@ -221,8 +222,10 @@ export class CalendarService extends Emitter {
                 const events = this._updateMeetingUrlOnEvents(formattedEvents);
 
                 if (!this._hasFetchedEvents
+                    || this._hasErrorOnLastFetch
```
paweldomas: that is getting to many flags in here, I'm trying to see if there's any way to simplify things

spot-client/src/spot-tv/calendars/calendarService.js
```diff
@@ -221,8 +222,10 @@ export class CalendarService extends Emitter {
                 const events = this._updateMeetingUrlOnEvents(formattedEvents);
 
                 if (!this._hasFetchedEvents
```
paweldomas: I don't understand why `_hasFetchedEvents` flag is needed exactly. Just don't see why comparing the last fetched events with the new ones is not enough?

spot-client/src/spot-tv/calendars/calendarService.js
paweldomas: If `_calendarEvents` would be undefined initially and on error(although I'm not sure if that's even necessary) then maybe the flag could be avoided?

virtuacoplenny: You won't let me stay lazy... Good thing there are unit tests. Hopefully they're true accurate. I don't remember what edge cases were trying to be worked around with the boolean but the tests pass so maybe that's enough.

PR #606, opened 2019-10-03
feat(in-call): show calendar event name if available

virtuacoplenny:
Screen Shot 2019-10-03 at 9 01 37 AM

Screen Shot 2019-10-03 at 8 50 58 AM

### DIFF
--- a/spot-client/src/spot-tv/ui/views/meeting.js
+++ b/spot-client/src/spot-tv/ui/views/meeting.js
@@ -115,7 +115,13 @@ export class Meeting extends React.Component {
      * @inheritdoc
      */
     render() {
-        const { invites, location, screenshare, startWithVideoMuted } = this._queryParams;
+        const {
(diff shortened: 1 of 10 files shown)


spot-client/src/spot-tv/ui/views/meeting.js
```diff
@@ -194,6 +201,7 @@ export class Meeting extends React.Component {
         const queryParams = new URLSearchParams(this.props.location.search);
         const invitesParam = queryParams.get('invites');
         const locationParam = queryParams.get('location');
+        const meetingDisplayNameParam = queryParams.get('meetingDisplayName');
```
paweldomas: Should it be URI decoded here?

virtuacoplenny: URLSearchParams does the decoding.

PR #607, opened 2019-10-04
feat(dial pad): remember selected country code

paweldomas:
If there's no country code provided by the backend, then restore the previously selected one.


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -307,9 +324,11 @@ function mapDispatchToProps(dispatch) {
  */
 function mapStateToProps(state) {
     const { phoneAuthorizeServiceUrl } = getSpotServicesConfig(state);
+    const countryCode = getCountryCode(state);
+    const mostRecentCountryCode = getMostRecentCountryCode(state);
 
     return {
-        countryCode: getCountryCode(state),
+        countryCode: countryCode || mostRecentCountryCode,
```
virtuacoplenny: should mostRecentCountryCode take precedence?

paweldomas: Yeah, at some point I was thinking the same. It's how Emil wanted it though. Also if you think more about it, Spot TV is not a personal device, but shared. So you may not want to call to the country another person was calling to in the meeting before.

virtuacoplenny: Is this prop more of a default country code then? I wonder if renaming it such would be better.

paweldomas: No, it is not default. It is used as default only if nothing's been configured on the backend.

PR #611, opened 2019-10-07
More i18n conversions

virtuacoplenny:


spot-client/src/common/ui/components/feedback/FeedbackOverlay.js
```diff
@@ -30,7 +31,8 @@ export function FeedbackOverlay(props) {
                     commentOnly = { true }
                     disableInactivityTimer = { true }
                     onSkip = { props.onCancelFeedback }
-                    onSubmitFeedback = { props.onSubmitFeedback } />
+                    onSubmitFeedback = { props.onSubmitFeedback }
+                    t = { props.t } />
```
paweldomas: why does it pass the `t` if `FeedbackFrom` is supposed to get it's own `t` "withTranslation"? Am I missing something?

virtuacoplenny: This one uses the FeedbackForm not hooked into redux.

PR #617, opened 2019-10-10
feat(spot-electron): specify product name

paweldomas:


spot-electron/package.json
```diff
@@ -1,6 +1,7 @@
 {
     "author": "Jitsi",
     "build": {
+        "productName": "Jitsi Spot",
```
saghul: Isn't it "Spot" that's how the mobile app is currently called.

paweldomas: Ok

PR #623, opened 2019-10-14
Integrate 'endpointId' returned by the backend service

paweldomas:


spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -217,13 +224,30 @@ export class SpotBackendService extends Emitter {
...
+                = registerDevice(`${this.pairingServiceUrl}`, pairingCode, storedEndpointId)
+                    .then(registration => {
+                        const { endpointId, ...rest } = registration;
+
+                        if (storedEndpointId !== endpointId) {
+                            logger.log('Storing new endpoint ID', {
+                                endpointId,
+                                endpointIdPersistenceKey: this.endpointIdPersistenceKey
+                            });
+                            persistence.set(this.endpointIdPersistenceKey, endpointId);
```
virtuacoplenny: what if we set and log each time no matter what? Could that be safer? Worse?

spot-client/src/common/backend/SpotBackendService.js
```diff
@@ -217,13 +224,30 @@ export class SpotBackendService extends Emitter {
         }
 
         if (!registerDevicePromise) {
+            const storedEndpointId = persistence.get(this.endpointIdPersistenceKey);
```
virtuacoplenny: Are there any cases where the endpointIdPersistenceKey needs to be cleared? The point of it is not to be cleared?

paweldomas: It seems that this one should not be cleared.

paweldomas: At least at this point I'm not aware of any cases when it should be cleared.

spot-client/src/common/backend/SpotBackendService.js
paweldomas: If register request is going to happen, then another message will be logged above which should contain the endpoint ID already(if it exists).
I'm not sure if we should store every time just for safety. Eventually it's been retrieved from storage already, so it it there?

PR #639, opened 2019-10-24
feat(analytics): report 'connection-failed' for Spot TV

paweldomas:


spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -200,26 +202,49 @@ export function createSpotTVRemoteControlConnection({ pairingCode, retry }) {
...
+            /**
+             * The connection is retried if explicit 'retry' flag has been set or if the connection succeeds initially.
+             * Even if both conditions are met, the connection will be dropped when using permanent pairing and
+             * the error is not recoverable.
+             *
+             * @type {boolean}
+             */
+            const willRetry
+                = (retry || initiallyConnected)
+                    && !(usingPermanentPairingCode && isUnrecoverableError);
```
virtuacoplenny: Would you be willing to break up these checks into other named variables? I'm having trouble keeping track of in what situations willRetry is true.

spot-client/src/app.js
```diff
@@ -34,6 +34,9 @@ import {
 import { SpotTVRemoteControlLoader } from './spot-tv/ui/loaders';
 import { SpotTvRestrictedRoute } from './spot-tv/routing';
 
+// This import enables Spot-TV specific analytics
+import 'spot-tv/analytics';
```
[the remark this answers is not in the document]
paweldomas: Import cycle, analytics was importing app-state and this line was importing analytics. Ideally this would be triggered from Spot TV entry point, but it's now shared with remote.

spot-client/src/spot-tv/app-state/actions.js
paweldomas: yeah... although I'm not sure how to name them. The ideal was that first line is one variable and the second one another. Do you have any suggestions?

virtuacoplenny: `canRecoverConnection` for the second one?

PR #640, opened 2019-10-24
Analytics for Spot Remote connection failed

paweldomas:


spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -145,14 +146,29 @@ export function connectToSpotTV(joinCode, shareMode) {
          * @returns {Promise}
          */
         function onDisconnect(error) {
-            logger.error('On Spot Remote disconnect', { error });
+            const usingPermanentPairingCode = getPermanentPairingCode(getState()) === joinCode;
```
virtuacoplenny: Does the permanent code not change? It's always the join code value even after code refreshes?

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -145,14 +146,29 @@ export function connectToSpotTV(joinCode, shareMode) {
...
+            const isUnrecoverableError = remoteControlClient.isUnrecoverableRequestError(error);
```
virtuacoplenny: Maybe rename to isRecoverableError? Maybe that should have been the method name this whole time and it never occurred to me.

paweldomas: yeah, but can it be a separate PR?

spot-client/src/spot-remote/app-state/actions.js
paweldomas: Yes, with the current code(that is pending refactor, that's why it's confusing) it does not change, because it's not really used as long as the backend has access/refresh token stored and valid.

paweldomas: Also the permanent code does not refresh really. Maybe you're thinking about long live remote pairing code.

PR #653, opened 2019-10-30
Tunnel electron logs to the spot-client logger

paweldomas:


spot-electron/src/spot-client-log-transport/SpotClientLogTransport.js
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Log levels supported by the jitsi logger.
+ */
+const levels = [ 'trace', 'debug', 'info', 'log', 'warn', 'error' ];
+
+const LOG_CACHE_SIZE = 1000;
+
+/**
+ * A class
```
virtuacoplenny: Yes

spot-electron/src/spot-client-log-transport/SpotClientLogTransport.js
```diff
@@ -0,0 +1,56 @@
...
+ */
+class SpotClientLogTransport {
+    /**
+     * Creates a new instance of {@code LoggingService} and registers itself
+     * to receive logs.
+     */
+    constructor() {
+        for (const level of levels) {
+            this[level] = this.log.bind(this);
```
virtuacoplenny: Is there a need to differentiate the different levels of logging?

spot-electron/src/spot-client-log-transport/SpotClientLogTransport.js
```diff
@@ -0,0 +1,56 @@
...
+     * The log message executed for every logging level.
+     *
+     * @param {...*} args - Log method arguments as defined by jitsi-logger.
+     * @returns {void}
+     */
+    log(...args) {
+        const logString = args.join(' ');
+
+        if (clientController.canSendClientMessage()) {
+            clientController.sendClientMessage('spot-electron-logs', logString);
```
virtuacoplenny: Do you think logging the context object will be needed soon?

paweldomas: yes

spot-electron/src/spot-client-log-transport/SpotClientLogTransport.js
paweldomas: yes

virtuacoplenny: I was not being clear. I meant right now only the level "log" gets logged. Do you also need to tell the client right now to send error or warn or other levels?

paweldomas: I think this should support all levels, yes

spot-electron/src/spot-client-log-transport/SpotClientLogTransport.js
virtuacoplenny: Let's never change this.

PR #654, opened 2019-10-30
ref(connection): show reconnect indicator instead of loading page 

virtuacoplenny:
tv in meeting
Screen Shot 2019-10-31 at 9 14 05 AM

tv at home
Screen Shot 2019-10-31 at 9 10 55 AM

remote control
Screen Shot 2019-10-31 at 9 10 49 AM


spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -192,8 +195,6 @@ export function connectToSpotTV(joinCode, shareMode) {
 
             _clearSubscriptions();
 
-            dispatch(clearSpotTVState());
```
paweldomas: why does it go away exactly? is it to not drop out of the remote control view? is there another place where this cleanup will happen?

virtuacoplenny: This is already called on line 171. I think this is a duplicate call to clearSpotTVState.

virtuacoplenny: Gonna add it back. If we want to remove it then we should do it at some other time. Keep this PR focused on the reconnect logic only.

spot-client/src/spot-remote/app-state/actions.js
```diff
@@ -139,6 +140,7 @@ export function connectToSpotTV(joinCode, shareMode) {
                 isSpotTv: false,
                 isPairingPermanent: backend && backend.isPairingPermanent()
             }));
+            dispatch(reconnectScheduleUpdate(false));
```
paweldomas: I am wondering if it would be better for safety to also do this call in `disconnectFromSpotTV`. It seems to be duplicating other cleanup logics as well.

spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -154,6 +155,7 @@ export function createSpotTVRemoteControlConnection({ pairingCode, retry }) {
             dispatch(setTenant(tenant));
             dispatch(setPermanentPairingCode(permanentPairingCode));
             roomProfile && dispatch(setCustomerId(roomProfile.customerId));
+            dispatch(reconnectScheduleUpdate(false));
```
paweldomas: Similar to remote, the TV also has explicit disconnect method which duplicates the cleanup in `disconnectSpotTvRemoteControl`. I would put this call for safety there as well. Or maybe test that scenario which might not be easy or maybe we dont do explicit disconnect yet while reconnecting anywhere in the app... so maybe just better put it there?

PR #658, opened 2019-11-04
feat(meeting): add a button to cancel a meeting join in progress

virtuacoplenny:
So if there is a slow join the user can immediately stop the
join.

PR-URL JIT-3413

Screen Shot 2019-11-04 at 3 23 59 PM

### DIFF
--- a/spot-webdriver/specs/adhoc-meeting.spec.js
+++ b/spot-webdriver/specs/adhoc-meeting.spec.js
@@ -19,6 +19,8 @@ describe('Can start a meeting', () => {
         const spotRemote = session.getSpotRemote();
         const inMeetingPage = spotRemote.getInMeetingPage();
 
+        // Go to an invalid meeting url to prevent any meeting loaded confirmation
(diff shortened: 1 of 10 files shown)


virtuacoplenny:
I should add a new selenium test for this.

spot-webdriver/specs/adhoc-meeting.spec.js
```diff
@@ -14,6 +14,19 @@ describe('Can start a meeting', () => {
         expect(spotTV.getMeetingName()).toBe(testMeetingName);
     });
 
+    it('and cancel join', () => {
+        const spotTV = session.getSpotTV();
+        const spotRemote = session.getSpotRemote();
+        const inMeetingPage = spotRemote.getInMeetingPage();
+
+        session.joinMeeting('https://meet.jit.si');
```
paweldomas: this locks, because it is not a valid meeting URL? maybe worth leaving a comment?

PR #659, opened 2019-11-05
Electron auto updater

paweldomas:


spot-client/src/common/auto-update/UpdateTimeRangeChecker.js
```diff
@@ -66,6 +63,7 @@ export class AutoUpdateChecker extends React.Component {
     componentWillUnmount() {
         logger.log('Stopping auto update checks');
 
+        this.props.onTimeWithinRangeUpdate(false);
```
virtuacoplenny: I'm wondering if it should be the responsibility of this component to emit false.

paweldomas: What else do you think should be doing that?

paweldomas: It seems to be doing what the name suggests, checking if the current time is within the time range. Those components are enabled when put into the DOM, so I was thinking need to emit false when being removed as in the auto updates are being disabled. Or should it be the view that controls `UpdateTimerRangeChecker`? I can also try moving this to a service class in a follow up PR if you want

virtuacoplenny: It might be a naming thing. onTimeWithinRangeUpdate can technically be true in that the current time is within a specified range, so to me setting it to false seems off, which makes me wonder if this is the place to do it. Again could be a naming thing and maybe the follow up would address that.

paweldomas: I don't know how to rename it better as that's what it does - checks if the current time is within the range. I was trying to move it to home and waiting-for-call views, but that creates more duplication. I agree this call doesn't fit in perfectly and may be misleading, but it's the least changes at this point.

PR #661, opened 2019-11-06
feat(analytics): report meeting summary

paweldomas:
meetingUrl - The full meeting URL.
duration - Duration of the meeting in seconds.
participantCount - How many participants were in the meeting(max over entire duration)


spot-client/src/spot-tv/ui/views/meeting.js
```diff
@@ -246,12 +247,23 @@ export class Meeting extends React.Component {
...
-            logger.log('Leaving meeting due to error', { error: leaveEvent.error });
-            this.props.onError(leaveEvent.errorCode, leaveEvent.error);
+        const {
+            error,
+            errorCode,
+            meetingSummary
+        } = leaveEvent;
+
+        if (meetingSummary) {
+            this.props.processMeetingSummary(meetingSummary);
```
virtuacoplenny: Should the summary include any errors?

spot-client/src/spot-tv/ui/components/meeting-frame/meeting-frame.js
```diff
@@ -541,12 +566,19 @@ export class MeetingFrame extends React.Component {
...
      */
     _onParticipantJoined({ id, displayName }) {
         this._participants.set(id, displayName);
+        const newMax = this._participants.size + 1;
+
+        if (this.state.maxParticipantCount < newMax) {
+            this.setState({
+                maxParticipantCount: this._participants.size + 1
+            });
+        }
```
virtuacoplenny: I don't think you need to set this on state. state is usually used for triggering a re-render, whereas these don't need to.

paweldomas: 🤦‍♂

PR #662, opened 2019-11-07
feat(electron): notarize Mac app

paweldomas:
Adds the post build scrip which executes 'electron-notarize' to notarize the Mac app.

### DIFF
--- a/spot-electron/package.json
+++ b/spot-electron/package.json
@@ -1,11 +1,21 @@
 {
     "author": "Jitsi",
     "build": {
+        "afterSign": "scripts/notarize.js",
         "productName": "Spot",
         "appId": "org.jitsi.spot",
         "mac": {
             "category": "public.app-category.business",
-            "target": "default"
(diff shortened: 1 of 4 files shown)


spot-electron/package.json
```diff
@@ -14,13 +24,17 @@
...
+        "electron-notarize": "0.1.1",
         "electron-rebuild": "1.8.5",
         "eslint": "5.16.0",
         "eslint-config-jitsi": "github:jitsi/eslint-config-jitsi#1.0.1",
         "eslint-plugin-flowtype": "3.10.1",
         "eslint-plugin-import": "2.17.3",
         "eslint-plugin-jsdoc": "7.2.3"
     },
+    "dmg": {
+        "sign": false
```
virtuacoplenny: Why false?

paweldomas: With the extra notarization signing step after build the signing happens there and not during regular build phase.

PR #664, opened 2019-11-08
Silent xmpp reconnect for some errors

virtuacoplenny:


virtuacoplenny:
Opening a little early to get feedback. I'll see about writing tests for this.

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -135,7 +135,8 @@ export class BaseRemoteControlService extends Emitter {
                     retryOnUnauthorized,
                     roomName: roomInfo.roomName,
                     roomLock: roomInfo.roomLock,
-                    onDisconnect: this._onDisconnect
+                    onDisconnect: this._onDisconnect,
+                    shouldAttemptReconnect: this._shouldXmppAttemptReconnect.bind(this)
```
paweldomas: please comment the new options in JS Docs

virtuacoplenny:
> This looks good. I guess now you need to cover the case when XMPP is reconnecting and P2P drops in the meantime plus we need tests to cover those corner cases.

Is the case not covered enough? Right now the logic before trying to reconnect, XMPP ask RCS if the P2P connection is active. If not active XMPP will say it has disconnected. In the case where XMPP is reconnect and p2p drops, XMPP should reach the disconnected logic at some point if there is continued failure.

paweldomas:
> Is the case not covered enough? Right now the logic before trying to reconnect, XMPP ask RCS if the P2P connection is active. If not active XMPP will say it has disconnected. In the case where XMPP is reconnect and p2p drops, XMPP should reach the disconnected logic at some point if there is continued failure.

I'm a bit afraid what will happen if the XMPP reconnect takes too long or gets stuck(it can when network is not responsive). I can see people trying to click buttons on remote and then coming back to us complaining that remote is not working.

paweldomas:
But we can also start with this and then improve if it turns out to be a problem.

spot-client/src/common/remote-control/BaseRemoteControlService.js
virtuacoplenny: Force pushed to add the js doc.

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -304,6 +304,15 @@ export default class XmppConnection {
             });
     }
 
+    /**
+     * Returns whetheer or not the internal XMPP connection is active.
+     *
+     * @returns {boolean}
+     */
+    isConnected() {
+        return this.xmppConnection?.xmpp.connection.connected;
```
paweldomas: Can you use a flag or something rather than reaching into the LJM internals?

virtuacoplenny: I decided to create a new flag to make the timing of when it gets set to true/false to be more accurate.

virtuacoplenny:
Force pushed as I rebased onto latest master to get some test fixes. Along with the force push is a new commit which fixes existing tests to account for the silent reconnect behavior, by explicitly killing p2p in those tests when the network is set offline.

I will now focus on some new tests to exercise silent reconnect.

spot-webdriver/specs/spot-tv-reconnect.spec.js
```diff
@@ -3,6 +3,24 @@ const spotSessionStore = require('../user/spotSessionStore');
...
+    function joinMeeting() {
+        const tv = session.getSpotTV();
+        const remote = session.getSpotRemote();
+
+        session.connectRemoteToTV();
+
+        session.joinMeeting();
+
+        remote.getInMeetingPage().waitForVisible();
+        tv.getMeetingPage().waitForMeetingJoined();
```
paweldomas: Why not move those waits to `session.joinMeeting` and remove this method? I just feel like this logic will be duplicated among other tests.

spot-webdriver/user/spot-user.js
```diff
@@ -105,13 +105,38 @@ class SpotUser {
...
         }, this._internalRemoteControlServiceName);
     }
 
+    /**
+     * Stops messages from being sent to or received from the signaling service.
+     *
+     * @returns {void}
+     */
+    stopSignalingConnection() {
+        this.driver.setNetworkConditions(_OFFLINE_NETWORK_CONDITIONS);
```
paweldomas: I would rather leave this call in `setNetworkOffline`, as this method might be used without turning the network offline

paweldomas: I would rather have the network offline call in the `setNetworkOffline` method, because that's how the method is called. This one could be renamed to 'waitForSignalingConnectionStopped`/`disconnected`.

spot-webdriver/user/spot-user.js
```diff
@@ -131,6 +156,30 @@ class SpotUser {
...
+                    const internalXmppConnection = window.spot[rcsName].xmppConnection;
+
+
+                    return internalXmppConnection.isConnected()
+                        && internalXmppConnection._hasJoinedMuc;
+                } catch (e) {
+                    return false;
+                }
+            }, this._internalRemoteControlServiceName),
+            120000,
```
paweldomas: This very long wait currently caused by the fact that it needs to wait for the BOSH session to timeout, right? Is it possible to reference a constant, so that we know which waits to change if we would be moving away from BOSH?

spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -55,6 +55,21 @@ export default class P2PSignalingBase extends Emitter {
         return this._peerConnections.get(remoteAddress);
     }
 
+    /**
+     * Finds if any connections are currently alive.
+     *
+     * @returns {boolean}
+     */
+    hasActiveConnection() {
+        for (const peerConnection of this._peerConnections.values()) {
```
paweldomas: Is it possible to use `find` on `.values()` ?

spot-webdriver/user/spot-user.js
virtuacoplenny: The test needs to stop signaling while keeping p2p alive. Now stop network kills both signaling and p2p. If stop network should continue to kill both, then another method will be needed to kill only signaling, which is how this stopSignalingConnection method came to be. How do you think signaling by itself should be killed?

spot-client/src/common/remote-control/P2PSignalingBase.js
virtuacoplenny: Not that I know of because that's an iterator, but I've made a change to turn the peerConnections map into an array so .find can be called on it.

spot-webdriver/specs/spot-tv-reconnect.spec.js
virtuacoplenny: I've moved any waits I found after joinMeeting() into joinMeeting().

spot-webdriver/user/spot-user.js
virtuacoplenny: The long waits are from absolute frustration from, yes, waiting for the BOSH session timeout. I'll add a constant called THIS_GHOST_PARTICIPANT_WONT_GO_AWAY or something.

spot-webdriver/specs/spot-tv-reconnect.spec.js
virtuacoplenny: I take that back. that introduced failing tests. I'll get back to this change.

spot-webdriver/user/spot-user.js
virtuacoplenny: Maybe the changes you want are there now. `setNetworkOffline` is back to only changing network conditions. I made a helper for the repeated killing of xmpp and p2p.

PR #667, opened 2019-11-11
ref(dial pad): detect the country even if no + was entered

paweldomas:
Improves country detection by using the parse function which detects country specific syntax. For example 00{country code} or 011{country code} in the US.

### DIFF
--- a/spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
+++ b/spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
@@ -1,4 +1,4 @@
-import { AsYouType } from 'libphonenumber-js';
+import { formatIncompletePhoneNumber, parsePhoneNumberFromString } from 'libphonenumber-js';
 import PropTypes from 'prop-types';
 import React from 'react';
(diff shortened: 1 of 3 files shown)


spot-client/src/spot-remote/ui/components/dial-pad/dial-pad.js
```diff
@@ -182,12 +207,11 @@ export class DialPad extends React.Component {
      */
     _onCountryCodeSelect(code) {
         this.setState({
-            selectedCountryCode: code,
+            defaultCountryCode: code,
+            parsedCountryCode: code,
```
virtuacoplenny: In this case we're getting the country code from the lib so we can trust its format?

paweldomas: What do you mean? I'm not sure what to change here

paweldomas: Do you want me to add a check if there's such country code? In this case it's coming from the country selector and not from the list, so it's predefined list.

PR #671, opened 2019-11-14
fix(controller): stop supporting iOS 11

virtuacoplenny:


saghul:
What issues does this solve? Given the amount of functionaility this provides (ie, not that much), one could imagine old devices could be repruposed as controllers. As it stands the 4th generation iPad would get the axe, no big deal sinec it's like 6 years old, but having some reasoning would be nice going forward.

virtuacoplenny:
I'm just gonna close this. I mainly made this to test slack notifications for pull requests.

PR #672, opened 2019-11-15
fix(docker build): clean before build

paweldomas:


scripts/docker-build.sh
```diff
@@ -8,6 +8,8 @@ fi
 DOCKER_REPO=103425057857.dkr.ecr.us-west-2.amazonaws.com/jitsi/spot-client
 
 cd spot-client
+rm -rf node_modules
```
saghul: `git clean -dfx` ?

paweldomas: is that a hard reset that can clean my local files? I'd rather not go that far, as we're doing those builds from local dev repo right now and I have custom scripts sitting there

saghul: 👍

PR #674, opened 2019-11-18
Auto update controller

paweldomas:


paweldomas:
I'm still testing/looking into adding tests.

spot-client/src/common/auto-update/AutoUpdateController.js
```diff
@@ -0,0 +1,113 @@
...
+        );
+
+        nightlyReloadTimePoller.start();
+
+        // Checks for Web updates
+        setInterval(() => {
+            if (!isWebUpdateAvailable(getState()) && isOkToUpdate(getState())) {
+                dispatch(checkForWebUpdateAvailable());
+            }
+        },
```
virtuacoplenny: It feels to me conceptually all of this logic in the middleware could be moved into a separate class and the middleware calls into the class to handle the logic. So in other words, much like what TimeRangePoller does but with even more stuff abstracted into TimeRangePoller. I'd rather keep the integration layer between the app and "internal libs" be as thin as possible. What do you think? My main concern with jitsi middleware has always been how so much logic gets put into it, eventually leading to being unable to know what gets called when, so centralizing logic might help mitigate that (as would not have auto-registration of middleware).

paweldomas:
Force pushed to include JIT ticket reference

PR #681, opened 2019-11-22
Fix/p2p corner cases

paweldomas:


spot-client/src/common/remote-control/P2PSignalingBase.js
```diff
@@ -325,19 +325,23 @@ export default class P2PSignalingBase extends Emitter {
...
+        if (connection && connection.isDataChannelActive()) {
+            logger.error('Received offer, but PeerConnection is still there', {
+                data,
+                from
+            });
 
-        if (!offer) {
             return;
+        } else if (connection) {
+            this.closeConnection(from);
```
virtuacoplenny: In this case the other side of the connection should already have closed its p2p? So this is more of a cleanup step for a dangling connection? Is there a chance that closing this connection could somehow signal the other side to close an active connection?

spot-client/src/common/remote-control/P2PReconnectTrait.js
```diff
@@ -0,0 +1,160 @@
...
+    /**
+     * Triggered whe signaling goes offline/online.
+     *
+     * @param {boolean} isConnected - The new signaling status.
+     * @returns {void}
+     */
+    _onSignalingConnectedUpdate(isConnected) {
+        isConnected && this._ensureP2PConnectionStarted();
+    }
+}
```
virtuacoplenny: I haven't heard of `trait` before. What does it mean?

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -563,12 +579,12 @@ export class RemoteControlClient extends BaseRemoteControlService {
     _onSpotTvJoinedMuc(spotTvAddress) {
         logger.log('Spot TV joined the MUC', { spotTvAddress });
 
-        // FIXME this doesn't check if the data channels is active. P2P reconnect logic will be added as next step.
-        if (this._p2pSignaling && !this._p2pSignaling.getConnectionForAddress(spotTvAddress)) {
-            // Client initiates the P2P signaling session
-            logger.log('Will try to establish P2P connection with Spot TV');
-            this._p2pSignaling.start(spotTvAddress);
+        if (this._p2pSignaling && !this._spotTvP2PRetry) {
+            this._spotTvP2PRetry = new P2PReconnectTrait(this._p2pSignaling, this.xmppConnection);
```
virtuacoplenny: What would you think of making a decorator on top of p2p that handles reconnect instead of having both _p2pSignaling and _spotTvP2PRetry?

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -239,6 +239,16 @@ export class BaseRemoteControlService extends Emitter {
...
     }
 
+    /**
+     * Returns the ICE servers to be used to establish the P2P signaling connection.
+     *
+     * @returns {?Array<RTCIceServer>}
+     * @private
+     */
+    _getP2PIceServers() {
+        return this.xmppConnection?.getJitsiConnection()?.xmpp?.connection?.jingle?.p2pIceConfig?.iceServers;
```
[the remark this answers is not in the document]
paweldomas: it was to not add public interface in LJM for that

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -487,6 +496,10 @@ export default class XmppConnection {
      * @returns {void}
      */
     _onDisconnect(error, reason) {
+        if (this._silentReconnectPromise) {
+            logger.error('Silent reconnect promise not cleared');
```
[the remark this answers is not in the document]
paweldomas: No, this is a case that is supposed to never happen. It's logged just to make debugging easier if code gets broken during refactoring or something

spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -501,17 +501,8 @@ export class RemoteControlClient extends BaseRemoteControlService {
...
-                    this._p2pSignaling.stop();
-                    this._p2pSignaling = null;
-                }
-                if (this._getBackend()) {
-                    // With backend it is okay for remote to sit in the MUC without Spot TV connected.
-                    this._resetSpotTvState();
-                } else {
-                    this._onDisconnect(CONNECTION_EVENTS.SERVER_DISCONNECTED);
-                }
+                this._spotTvInMuc = false;
```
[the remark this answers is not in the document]
paweldomas: I was thinking about spot status, but this is preserved/left untouched when TV leaves the MUC, but P2P is still active.

spot-client/src/common/remote-control/P2PReconnectTrait.js
paweldomas: Well Im not english specialist so dont ask me lol. In this case it's kind of a feature enabled on another feature, which is doing reconnects on broken p2p session. Maybe it could be renamed to session. Was just trying to make it more about the reconnect.

spot-client/src/common/remote-control/remoteControlClient.js
paweldomas: So this decorator would wrap the P2P signaling and add extra methods?

spot-client/src/common/remote-control/P2PReconnectTrait.js
paweldomas: I think I took the word from scala: https://www.tutorialspoint.com/scala/scala_traits.htm it is used there as another way of calling interface, that has some implementation. Maybe this doesnt make sense. I can just call it reconnect decorator or something around those lines if we decide to make it a decorator?

spot-client/src/common/remote-control/remoteControlClient.js
virtuacoplenny: Yeah. p2pSignaling remains the same, just like it is now in the PR, but a new class will take in p2pSignaling (or instantiate it itself) and provide additional methods or functionality on top of p2pSignaling, and only this new class is used by RCS. That way conceptually all the p2p logic should be in this one new class. There will be boring work to do to pass through method calls though. I guess you could also extend the p2pSignaling class to get the same effect. These are just some alternatives I'm pondering right now; I'm not recommending that you actually make changes.

paweldomas: Okay, then I got much less excited about the decorator when I heard about `boring work to do to pass through method calls`. Not really sure what benefit would that give other than one less field in RCS.

paweldomas: If you would prefer one class then it can be as well moved over to the P2PSignalingClient?

paweldomas: The initial reason why I've put it into separate was to avoid piling logics in one file, making it aware of "signaling connection connected/disconnected".

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -314,6 +314,15 @@ export default class XmppConnection {
         return this._isXmppConnectionActive;
     }
 
+    /**
+     * Returns whether or not the internal XMPP connection is being reconnected.
+     *
+     * @returns {boolean}
+     */
+    isReconnecting() {
+        return Boolean(this._silentReconnectPromise) || Boolean(this.initPromise);
```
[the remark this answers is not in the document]
paweldomas: Im thinking that `this.initPromise` should be reset after the join Muc process is done then I could re-use it for that purpose.

On a side note it doesn't seem like a correct behavior to return this promise because the options could be different and in general joinMuc should not be called without leaving the Muc first.

spot-client/src/common/remote-control/remoteControlClient.js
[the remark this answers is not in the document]
paweldomas: I've added this to the constructor, as spot state can not be really reused for that.

spot-client/src/common/remote-control/xmpp-connection.js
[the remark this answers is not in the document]
paweldomas: I've pushed a follow up commit https://github.com/jitsi/jitsi-meet-spot/pull/681/commits/16a906bdfb43f007e76d80e3d5e6a1aa84f5c4ec please let me know WDYT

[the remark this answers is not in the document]
paweldomas: addressed in https://github.com/jitsi/jitsi-meet-spot/pull/681/commits/49e3253c271c0a14389e39d929df338e54c363a0

spot-client/src/common/remote-control/xmpp-connection.js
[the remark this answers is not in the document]
paweldomas: attempt to address in: https://github.com/jitsi/jitsi-meet-spot/pull/681/commits/16a906bdfb43f007e76d80e3d5e6a1aa84f5c4ec

spot-client/src/common/remote-control/P2PSignalingBase.js
paweldomas: Yes it is a cleanup for dangling connection. It should have already closed it otherwise would not be able to send a new offer. Anyway on that other side it would get "disconnected" ICE state when it's closed here, but it should not really happen.

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -119,8 +127,13 @@ export default class XmppConnection extends Emitter {
         } = options;
 
         if (this.initPromise) {
+            logger.error('joinMuc called before another joinMuc job completed');
+
             return this.initPromise;
         }
+        if (this._isXmppConnectionActive) {
+            logger.error('joinMuc called while another connection is still active');
```
virtuacoplenny: Maybe it should throw?

paweldomas: maybe it should, I can do this change if you like... I was just hoping that the app might be still working to some extend, but would see in the logs that things are messed up

paweldomas: I will change those log error to throws. We should be able to catch those during testing for the most part.

spot-client/src/common/remote-control/BaseRemoteControlService.js
```diff
@@ -87,6 +87,9 @@ export class BaseRemoteControlService extends Emitter {
             onPresenceReceived: this._onPresenceReceived
         });
 
+        !this._p2pSignaling || logger.error('Leaked P2PSignaling instance');
```
virtuacoplenny: Do you not want to actively prevent the leak?

spot-client/src/common/remote-control/remoteControlClient.js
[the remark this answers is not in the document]
virtuacoplenny: I noticed XMPP connection provides a `type` field that indicates a join. Could that be used?

spot-client/src/common/remote-control/BaseRemoteControlService.js
paweldomas: It is preventing the leak by doing the cleanup. It is just a debug log message to catch the corner case and save time on debugging in case the cleanup is broken.

paweldomas: Oh you mean to do the cleanup when it happens, no because someone could think it is a supported case.

PR #687, opened 2019-11-26
feat(meeting): temporary flag for zoom support

virtuacoplenny:
So zoom-related code can be committed and tested
manually without exposing to users.


virtuacoplenny:
It'll be easier at this point to merge this change in with my work in progress.

PR #688, opened 2019-11-26
ref(meeting): pass along what type of meeting is being joined

virtuacoplenny:
To differentiate between jitsi and other providers so that
the appropriate UI can be displayed.

### DIFF
--- a/spot-client/src/common/app-state/remote-control-service/actions.js
+++ b/spot-client/src/common/app-state/remote-control-service/actions.js
@@ -75,12 +75,20 @@ export function joinWithScreensharing(meetingName, screensharingType) {
  * @param {string} meetingName - The Jitsi meeting to join.
  * @param {string} meetingDisplayName - An alternative display name for the
  * meeting, such as the event name on a calendar.
+ * @param {string} meetingType - The meeting provider which should host the meeting
+ * of the provided meetingName.
  * @returns {Function}
  */
-export function joinScheduledMeeting(meetingName, meetingDisplayName) {
+export function joinScheduledMeeting(meetingName, meetingDisplayName, meetingType = 'jitsi') {
     return dispatch => createAsyncActionWithStates(
         dispatch,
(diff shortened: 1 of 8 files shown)


virtuacoplenny:
What about if instead of type, for zoom meetings the remote passes the entire zoom url. And then the TV can figure out the type based on the url? So anything with zoom.us in it would be meeting type zoom but all else would be assumed jitsi?

paweldomas:
`What about if instead of type, for zoom meetings the remote passes the entire zoom url. And then the TV can figure out the type based on the url? So anything with zoom.us in it would be meeting type zoom but all else would be assumed jitsi?`

I guess that could also work as long as we'll be consistent on always passing the full URL.

virtuacoplenny:
We might not be consistent. Due to the tenant name ,the assumption was the TV would take care of that when only the meeting name is passed to the TV from the remote. I think intentionally we wanted the TV to take care of that?

paweldomas:
> We might not be consistent. Due to the tenant name ,the assumption was the TV would take care of that when only the meeting name is passed to the TV from the remote. I think intentionally we wanted the TV to take care of that? Edit: but maybe what we really do want is for the remote to always send the full URL, even for jitsi meetings.

I think we may be better off with always passing full URL instead of doing shananigans around tenant. If there will be thing that remote needs to get from the tv to craft the URL we can pass that in the TV state.

paweldomas:
> I think we may be better off with always passing full URL instead of doing shananigans around tenant.

Btw. I think recently there was change to pass the full url for the tenant case.

paweldomas:
I was thinking about this commit: https://github.com/jitsi/jitsi-meet-spot/commit/db88101553f41656b8c4ad077c1ef522278ce86e

Looks like at least from the method names it's still possible to pass just the name.

virtuacoplenny:
Okay. I'm gonna close this PR and have the remote always pass the full url to the tv.

PR #693, opened 2019-12-02
SpotTV Electron exit password

paweldomas:


spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -325,18 +330,15 @@ export function createSpotTVRemoteControlConnection({ pairingCode, retry }) {
...
  * @param {string} [permanentPairingCode] - This one is optional and used only in the backend scenario where Spot TV
  * needs a permanent pairing code in order to connect to the service.
  * @returns {Promise<Object>} Resolves with an object containing the latest
  * join code and jwt.
  */
-function createConnection(state, permanentPairingCode) {
+function createConnection(state, backend, permanentPairingCode) {
     const joinCodeRefreshRate = getJoinCodeRefreshRate(state);
     const remoteControlConfiguration = getRemoteControlServerConfig(state);
-    const backend
```
virtuacoplenny: Why pass it in? I'm not against the change but wonder why you chose to do it.

spot-client/src/spot-tv/backend/SpotTvBackendService.js
```diff
@@ -70,6 +71,32 @@ export class SpotTvBackendService extends SpotBackendService {
...
+     */
+    fetchExitPassword() {
+        if (!this._getExitPasswordPromise) {
+            this._getExitPasswordPromise = fetchRoomInfo(
+                this.roomKeeperServiceUrl,
+                this.getJwt()
+            ).then(({ endpointPassword }) => {
+                this._cachedExitPassword = endpointPassword && endpointPassword.length ? endpointPassword : undefined;
+            }, error => {
+                logger.error('Fetch exit password failed', { error });
```
virtuacoplenny: Do you need to clear _getExitPasswordPromise? Is it really necessary to gate the execution of `fetchExitPassword` with `this. _getExitPasswordPromise`?

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
```diff
@@ -3,20 +3,98 @@ import React from 'react';
...
+        this._onInputChanged = this._onInputChanged.bind(this);
+    }
+
+    /**
+     * Shows the loading screen while the exit password is being checked.
+     *
+     * @returns {void}
+     */
+    componentDidMount() {
+        fetchExitPassword()
```
virtuacoplenny: I wonder if it should fetch on password on submit as well. That'll make sure to get a more recent password. However, I cannot really see it being a practical issue because someone in IT probably isn't frequently changing the exit password.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
```diff
@@ -3,20 +3,98 @@ import React from 'react';
...
+                        exitPassword
+                    });
+                } else {
+                    this.setState({
+                        viewToDisplay: VIEW_EXIT_CONFIRMATION
+                    });
+                }
+            }, () => {
+                this.props.onSomethingWentWrong();
+                this.props.onCancel();
```
virtuacoplenny: I'm not sure about the UX of preventing close if there is an error fetching the password. Maybe it should show the exit confirmation instead? That makes the password a bit less safe but the feature already has workarounds.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
```diff
@@ -26,18 +104,65 @@ export class ExitVerification extends React.Component {
...
+                        qaId = 'exit-electron-confirm' >
+                        { t('buttons.confirm') }
+                    </Button>
+                </StatusOverlay>
+            );
+        case VIEW_LOADING:
+            return (
+                <StatusOverlay
+                    qaId = 'exit-electron-loading'
+                    title = { t('appStatus.pleaseWait', { productName }) }>
```
virtuacoplenny: Maybe `productName` is unnecessary for this string.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.test.js
```diff
@@ -0,0 +1,130 @@
...
+    beforeEach(() => {
+        store = createStore((state = { config: {} }) => state);
+        onCancel = jest.fn();
+        onPasswordInvalid = jest.fn();
+        onVerification = jest.fn();
+    });
+
+    const mountExitComponent = () => {
+        exitComponent = mount(
+            <Provider store = { store }>
```
virtuacoplenny: I think don't the store is needed. I think if you use the non-connected ExitVerification then all the values can be passed in manually. Unless that is your wanting to test the full integration with redux.

spot-client/src/spot-tv/app-state/actions.js
paweldomas: I've tried to explain it in the commit message. The scope change is needed to preserve the exit password cache when internet is lost and the reconnect is in progress.

spot-client/src/spot-tv/backend/SpotTvBackendService.js
paweldomas: The intention was to prevent queuing multiple requests, but it looks like the UI ended up not allowing that. I will revert it.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
paweldomas: Oh yeah actually this argument is not used anymore 🤦‍♂

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.test.js
paweldomas: Right, I will try without the store.  It's using disconnected version anyway.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
paweldomas: The password is being fetched every time the x sign is clicked. So it should be very rare to hit this corner case when someone changes password while on the password prompt. Even so I would expect it natural thing to close the dialog and try again.

paweldomas: The promise was being cleared in the last `.then()` which is the common code executed for either resolve or reject.

spot-client/src/spot-tv/backend/SpotTvBackendService.js
paweldomas: The promise was being cleared in the last .then() which is the common code executed for either resolve or reject.

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.js
paweldomas: This is not UX for failed fetch password, but for the case that is supposed to never happen. The fetchExitPassword should never reject because on failed request will return 'undefined' or the most recently cached value. Maybe it should throw and crash the app instead?

paweldomas: I will change that to throw

spot-client/src/spot-tv/ui/components/exit-electron/ExitVerification.test.js
paweldomas: Turns out it is need for some of the nested components: `Background`

virtuacoplenny: What if you did a shallow mount?

virtuacoplenny: Nevermind, I think that might screw up the buttons.

paweldomas:
force pushed to squash the fixes

PR #694, opened 2019-12-03
enable android webview for spot TV

zbettenbuk:
Enable rendering Spot TV in Android WebViews for custom integrations


virtuacoplenny:
Do the features work okay on android? Like screensharing?

paweldomas:
Is it not a specific device being targeted? I've heard the plan was to open in a webview app and this app should add something extra to the user agent and enable it only to that app.

PR #695, opened 2019-12-04
feat(zoom): meeting number signing service

virtuacoplenny:


spot-integrations-service/src/controllers/ZoomSigningController.js
```diff
@@ -0,0 +1,62 @@
...
+     * @param {string} apiSecret - The Zoom api secret associated with a client
+     * integration application.
+     */
+    constructor(apiSecret) {
+        this._apiSecret = apiSecret;
+
+        // eslint-disable-next-line new-cap
+        this.router = express.Router();
+
+        this.router.post(routes.zoom.sign, this._onSignMeetingNumber.bind(this));
```
virtuacoplenny: I forgot to add payload validation middleware.

spot-integrations-service/README.md
```diff
@@ -0,0 +1,12 @@
+# Spot Integrations Service
+
+A backend service for signing Zoom meeting numbers with an api secret and api key.
+
+## Getting Started
+1. `npm install`
+1. `npm start`
+
+## Building
+1. `chmod +x ./scripts/build.sh`
```
paweldomas: is this step really required? why not include executable in git?

virtuacoplenny:
@paweldomas I'm wondering now if this should be a separate server from admin. What if this were combined with spot-admin instead? We had been talking about building out spot-admin anyway.

spot-integrations-service/README.md
virtuacoplenny: I guess then I already pushed it as executable. I don't know how file permission persistence works. I'll update the readme.

PR #698, opened 2019-12-05
feat: detect Spot TV conflict

paweldomas:
If Spot TV will fail to join the MUC with conflict error, for longer
than the max client idle timeout, it will mean that most likely another
Spot TV instance is already connected.

The conflict can not be detected instantly, because not gracefully
closed connections are cleaned up only after the idle timeout on
the server.

### DIFF
--- a/spot-client/src/common/i18n/en.json
+++ b/spot-client/src/common/i18n/en.json
@@ -31,6 +31,7 @@
         "pleaseWait": "Please wait",
         "reconnecting": "Reconnecting",
         "setup": "Currently in setup.",
+        "tvConflict": "Please make sure another window is not open and try again.",
         "tvNotSupported": "Hosting {{ productName }} is not supported on the current browser.",
         "unexpectedError": "An unexpected error occurred.",
         "useChrome": "Please open {{ productName }} on Chrome desktop.",
@@ -44,6 +45,7 @@
         "go": "Join",
         "select": "Select",
         "next": "Next",
+        "retry": "Retry",
         "send": "Send",
         "skip": "Skip",
         "submit": "Submit"
--- a/spot-client/src/common/i18n/en.json
+++ b/spot-client/src/common/i18n/en.json
@@ -31,7 +31,7 @@
         "pleaseWait": "Please wait",
         "reconnecting": "Reconnecting",
         "setup": "Currently in setup.",
-        "tvConflict": "Please make sure another window is not open and try again.",
+        "tvConflict": "An instance of {{ roomName }} is already open. Please close all other instances and try again.",
         "tvNotSupported": "Hosting {{ productName }} is not supported on the current browser.",
         "unexpectedError": "An unexpected error occurred.",
         "useChrome": "Please open {{ productName }} on Chrome desktop.",
--- a/spot-client/src/common/remote-control/xmpp-connection.js
+++ b/spot-client/src/common/remote-control/xmpp-connection.js
@@ -8,6 +8,15 @@ import { getJitterDelay } from 'common/utils';
 import { IQ_NAMESPACES, IQ_TIMEOUT } from './constants';
 
 /**
+ * How long will the join MUC function retry on conflict error before giving up.
+ * The value must be bigger than the idle client connection timeout. It's currently set to 30 seconds for the worst case
+ * scenario.
+ *
+ * @type {number}
+ */
+const CONFLICT_TIMEOUT = 35000;
+
+/**
  * XML element name for spot status added to MUC presence.
  *
  * @type {string}
@@ -214,6 +223,8 @@ export default class XmppConnection extends Emitter {
         );
 
         const createJoinPromise = function () {
+            let firstConflictTimestamp;
+
             return new Promise((resolve, reject) => {
                 const { connection } = this.xmppConnection.xmpp;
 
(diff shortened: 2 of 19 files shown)


spot-client/src/common/i18n/en.json
```diff
@@ -31,6 +31,7 @@
         "pleaseWait": "Please wait",
         "reconnecting": "Reconnecting",
         "setup": "Currently in setup.",
+        "tvConflict": "Please make sure another window is not open and try again.",
```
virtuacoplenny: Maybe this should be more specific and call out to check if another instance of the same Spot TV is open? And that only one can be open at a time?

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -7,6 +7,15 @@ import { getJitterDelay } from 'common/utils';
...
 import { IQ_NAMESPACES, IQ_TIMEOUT } from './constants';
 
+/**
+ * How long will the join MUC function retry on conflict error before giving up.
+ * The value must be bigger than the idle client connection timeout. It's currently set to 30 seconds for the worst case
+ * scenario.
+ *
+ * @type {number}
+ */
+const CONFLICT_TIMEOUT = 35000;
```
virtuacoplenny: Is there access to the idle client connection timeout? Maybe this value can be set by adding to that idle client value?

spot-client/src/common/i18n/en.json
paweldomas: Can you give me the exact text you have in mind? it will be easier this way

spot-client/src/common/remote-control/xmpp-connection.js
paweldomas: No, this value is defined in the prosody plugin in another repo.

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -230,6 +241,21 @@ export default class XmppConnection extends Emitter {
...
+                        if (error === 'conflict') {
+                            if (!firstConflictTimestamp) {
+                                firstConflictTimestamp = Date.now();
+                            }
+
+                            const timeSinceFirstConflict = Date.now() - firstConflictTimestamp;
+
+                            if (timeSinceFirstConflict <= CONFLICT_TIMEOUT) {
+                                logger.warn('Retry MUC join on conflict error', { timeSinceFirstConflict });
+                                setTimeout(() => this._joinMuc(roomLock), 5000);
```
virtuacoplenny: Maybe we should clear this timeout on destroy, to be safe?

spot-webdriver/specs/spot-tv-conflict.spec.js
```diff
@@ -0,0 +1,27 @@
...
+        const conflictPage = spotTv2.getConflictPage();
+
+        conflictPage.waitForVisible();
+
+        spotTv1.cleanup();
+
+        conflictPage.clickRetry();
+
+        spotTv2.getCalendarPage().waitForVisible();
+    });
```
virtuacoplenny: Does this need gating so the test runs only with the backend flow? I think in the open source flow there's a random room name generated.

spot-client/src/common/remote-control/xmpp-connection.js
paweldomas: yes, good point

spot-webdriver/page-objects/conflict-page.js
```diff
@@ -0,0 +1,44 @@
...
+const CONFLICT_VIEW = '[data-qa-id=conflict-view]';
+
+const RETRY_BUTTON = '[data-qa-id=conflict-retry';
+
+/**
+ * A page object for interacting with the conflict view of Spot-TV.
+ */
+class ConflictPage extends PageObject {
+    /**
+     * Initializes a new {@code CalendarPage} instance.
```
virtuacoplenny: CalendarPage -> ConflictPage.

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -544,8 +570,7 @@ export default class XmppConnection extends Emitter {
 
         if ((error === 'connection.droppedError'
             || error === 'connection.otherError'
-            || error === 'item-not-found'
-            || error === 'conflict')
+            || error === 'item-not-found')
```
virtuacoplenny: This is an interesting case. I wonder in what practical cases conflict can be triggered here.

spot-client/src/common/i18n/en.json
virtuacoplenny: Oh can we get the room name by this point? What if we do something like `{{ roomName}} is already running elsewhere. Please close the running instance and try again.`

spot-client/src/common/remote-control/xmpp-connection.js
paweldomas: Right now only ever I saw the 'conflict' triggered is when trying to join MUC with the same nickname.

In general it's safer to handle less cases here than more, because it is less risk of ending up in a broken loop.

If you mean when conflict can happen in real life then the cases I've found are:
- the case when another Spot TV is already in the MUC
- the case when network goes down, there's only ghost connection left. This one looks dangerous, but when network goes down the error will be different so the xmpp connection will go into the retry loop and when it hits the conflict on join MUC it will keep trying until after the idle timeout

I haven't found any other cases. It would have to be some other "conflict" returned not on the join MUC level and in such case it's better to break the loop for safety.

spot-client/src/common/i18n/en.json
paweldomas: We should have the room name. I'll try to add that.

spot-webdriver/specs/spot-tv-conflict.spec.js
paweldomas: yes

PR #699, opened 2019-12-06
ref(backend): remove unused 'emitted' field

paweldomas:
Once 'emitted' and 'expiresIn' gets converted to 'expires', then the 'emitted' field was never used for anything once the response gets outside of the util module.


paweldomas:
Merged as part of another PR

PR #704, opened 2019-12-08
fix(adhoc): prevent meeting join for non-whitelisted domains 

virtuacoplenny:
Screen Shot 2019-12-08 at 1 22 28 PM


virtuacoplenny:
Looks like I'll have to figure out how to fix the selenium tests to check the cancel-join button. Probably while I'm at it I can write a new selenium test for the changes in this pr.

spot-client/src/common/utils/meeting.test.js
```diff
@@ -0,0 +1,28 @@
+import { findWhitelistedMeetingUrl } from './meeting';
+describe('meeting utils', () => {
+    describe('findWhitelistedMeetingUrl', () => {
```
paweldomas: any tests that would cover the exclude pattern?

virtuacoplenny: Do you know why the exclude pattern even exists? I can't remember.

virtuacoplenny: The jitsi dial in info page...that's why.

PR #708, opened 2019-12-10
fix(zoom): prompt for password

virtuacoplenny:
Note that the zoom's modal is being dismissed before any re-join call is made. Currently clicking "OK" dismisses the modal and leaves the page open, which may or may not be intentional on zoom's part. With the modal dismissed, join can be retried without a timeout error occurring.

### DIFF
--- a/spot-client/src/common/zoom/errorCodes.js
+++ b/spot-client/src/common/zoom/errorCodes.js
@@ -1,0 +1,6 @@
+export default {
+    FAIL: 1,
+    ERROR_NOT_EXIST: 3001,
+    MEETING_NOT_START: 3008,
+    WRONG_MEETING_PASSWORD: 3004
+};
--- a/spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
+++ b/spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
@@ -6,7 +6,7 @@ import { getZoomConfiguration, setSpotTVState } from 'common/app-state';
 import { logger } from 'common/logger';
 import { COMMANDS, SERVICE_UPDATES } from 'common/remote-control';
 import { parseMeetingUrl } from 'common/utils';
-import { events } from 'common/zoom';
+import { errorCodes, events } from 'common/zoom';
 
 import { ApiHealthCheck } from '../ApiHealthCheck';
 import meetingFramePropTypes from '../meetingFramePropTypes';
@@ -155,6 +155,12 @@ export class ZoomMeetingFrame extends React.Component {
         case COMMANDS.SET_VIDEO_MUTE:
             this._zoomIframeManager.setVideoMute(data.mute);
             break;
+
+        case COMMANDS.SUBMIT_PASSWORD:
+            this.props.updateSpotTvState({ needPassword: false });
+            this._zoomIframeManager.submitPassword(data);
+
+            break;
         }
     }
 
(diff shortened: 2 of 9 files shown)


spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
```diff
@@ -187,21 +193,27 @@ export class ZoomMeetingFrame extends React.Component {
...
             const zoomErrorCode = data?.error?.code;
 
-            if (zoomErrorCode === 1 || zoomErrorCode === 3001) {
-                // The error code 1 is defined as a general failure in [1], but according to the internets[2] and local
-                // testing it can happen is other cases as well including when trying to join invalid meeting ID.
+            if (zoomErrorCode === errorCodes.FAIL || zoomErrorCode === errorCodes.ERROR_NOT_EXIST) {
+                // The error code FAIL is defined as a general failure in [1], but according to the internets[2] and
+                // local testing it can happen is other cases as well including when trying to join invalid meeting ID.
                 // [2]: https://devforum.zoom.us/t/joining-fail-with-error-code-1/6417
+                logger.warn('Failed to join zoom meeting', { zoomErrorCode });
```
paweldomas: this one can be removed as currently is logged at 218

spot-client/src/zoom/iframe-api/IframeApi.js
```diff
@@ -76,6 +76,16 @@ export class IFrameApi {
 
             break;
         }
+
+        case commands.SUBMIT_PASSWORD: {
+            sdk.submitPassword(data.password)
+                .then(
+                    () => this._sendMessageToParent(events.MEETING_JOIN_SUCCEEDED),
+                    error => this._sendMessageToParent(events.MEETING_JOIN_FAILED, { error })
```
paweldomas: what is the `error` in this case? could it be anything?

spot-client/src/zoom/sdk/sdk.js
```diff
@@ -177,6 +192,21 @@ class Sdk {
...
     }
 
+    /**
+     * Tries to rejoin the meeting but with a password.
+     *
+     * @param {string} passWord - The password to use for joining the meeting.
+     * @returns {void}
+     */
+    submitPassword(passWord) {
+        this._cachedJoinOptions = {
```
paweldomas: this might be not necessary as `this.joinMeeting` will overwrite it, so maybe just pass the new options?

spot-client/src/common/zoom/errorCodes.js
```diff
@@ -0,0 +1,6 @@
+export default {
+    FAIL: 1,
+    ERROR_NOT_EXIST: 3001,
+    MEETING_NOT_START: 3008,
```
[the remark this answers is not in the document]
virtuacoplenny: This one I will want to keep `MEETING_NOT_START` because in the zoom docs the constants is `MEETING_NOT_START`.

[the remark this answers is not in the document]
paweldomas: oh, really? okay

spot-client/src/zoom/iframe-api/IframeApi.js
virtuacoplenny: It is the reject from sdk.joinMeeting. It's usually the errorCode on failed meeting join, but it could be an error from something else in the join chain.

paweldomas: I see that it's ok because there's a generic error handler

PR #710, opened 2019-12-11
ref: auto whitelist zoom domain if enabled

paweldomas:
Automagically includes 'zoom.us' domain if zoom is enabled. Check for zoom enabled only on the redirect to meeting action, because it's earlier and seems to be the gate for the TV to join a meeting.

### DIFF
--- a/spot-client/src/spot-tv/app-state/actions.js
+++ b/spot-client/src/spot-tv/app-state/actions.js
@@ -8,6 +8,7 @@ import {
     getRemoteControlServerConfig,
     getSpotServicesConfig,
     getTemporaryRemoteIds,
+    isZoomEnabled,
     reconnectScheduleUpdate,
     removePairedRemote,
     setCustomerId,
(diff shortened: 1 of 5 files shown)


spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -535,12 +536,17 @@ export function updateSpotTVSource() {
  */
 export function redirectToMeeting(meetingNameOrUrl, { invites, meetingDisplayName, screenshare, startWithVideoMuted }) {
     return (dispatch, getState) => {
+        const state = getState();
+        const domainWhitelist = getMeetingDomainsWhitelist(state);
+
+        isZoomEnabled(state) && domainWhitelist.push('zoom.us');
```
virtuacoplenny: Maybe `.*zoom.us`? They support vanity urls.

paweldomas: can you give any example? I'm afraid `.*zoom.us` will include malicious URLs like `123zoom.us`

paweldomas: `.*.zoom.us` ?

virtuacoplenny: ah crap you're right. they support urls like https://1234.zoom.us

virtuacoplenny: This makes me wonder if we shouldn't be using regex at all. Instead we should take the domain and split it on period and match based on that? Just do some really dumb, generic string matching.

paweldomas: I like the ends in .domain idea. I will try to take a look at least.

PR #711, opened 2019-12-11
fix(zoom): retry to join meeting that has not started

virtuacoplenny:
Screen Shot 2019-12-11 at 3 14 31 PM

### DIFF
--- a/spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
+++ b/spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
@@ -34,6 +34,7 @@ export class ZoomMeetingFrame extends AbstractMeetingFrame {
         super(props);
 
         this._zoomIframeManager = null;
+        this._joinRetryTimeout = null;
 
         this._rootRef = React.createRef();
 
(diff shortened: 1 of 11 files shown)


spot-client/src/spot-tv/ui/components/meeting-frame/ZoomMeetingFrame/ZoomMeetingFrame.js
```diff
@@ -185,14 +217,16 @@ export class ZoomMeetingFrame extends AbstractMeetingFrame {
...
+
+                this.props.updateSpotTvState({ waitingForMeetingStart: true });
+                this._enqueueJoinRetry();
             } else if (zoomErrorCode === errorCodes.WRONG_MEETING_PASSWORD) {
                 logger.log('password required');
 
-                this.props.updateSpotTvState({ needPassword: true });
+                this.props.updateSpotTvState({
+                    needPassword: true,
+                    waitingForMeetingStart: true
```
paweldomas: why does the password case also have `waitingForMeetingStart` set to `true`?

virtuacoplenny: Oh good point. It should be false. Makes me wonder what I was testing when I made this change... Probably the waitingForMeetingStart  state got cleared on meeting join so I wasn't testing anything.

virtuacoplenny: Yes, I definitely wanted this to be false. it hides the waiting overlay.

PR #718, opened 2019-12-14
ref(zoom): use a combined selector to calculate domain whitelist

virtuacoplenny:
This way multiple code paths can use the same logic to get
the full list of whitelisted domains,, which is now a generated
list due to zoom being whitelisted without being explicitly
specified as whitelisted in the config.

Reselect is brought in so that the generated array
of allowed domains can be memoized to avoid triggering
re-renders when the array is passed in as props.

### DIFF
--- a/spot-client/src/spot-tv/app-state/actions.js
+++ b/spot-client/src/spot-tv/app-state/actions.js
@@ -4,11 +4,10 @@ import {
     clearAllPairedRemotes,
     destroyConnection,
     getJoinCodeRefreshRate,
-    getMeetingDomainsWhitelist,
+    getAllAllowedMeetingDomains,
     getRemoteControlServerConfig,
     getSpotServicesConfig,
     getTemporaryRemoteIds,
-    isZoomEnabled,
     reconnectScheduleUpdate,
     removePairedRemote,
     setCustomerId,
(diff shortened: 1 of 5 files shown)


spot-client/src/spot-tv/app-state/actions.js
```diff
@@ -537,9 +536,7 @@ export function updateSpotTVSource() {
 export function redirectToMeeting(meetingNameOrUrl, { invites, meetingDisplayName, screenshare, startWithVideoMuted }) {
     return (dispatch, getState) => {
         const state = getState();
-        const domainWhitelist = getMeetingDomainsWhitelist(state);
```
paweldomas: I'm wondering if appending 'zoom.us' at the `getMeetingDomainsWhitelist` level, would make things simpler?

virtuacoplenny: One concern I had was always creating a new array, which could trigger a re-render if it's being passed in directly. I also worried about mutating the array and if that's unexpected. So I though this reselect approach would be nice. But I could munge the config while we're setting the default config. what do you think of that? So instead of at the `getMeetingDomainsWhitelist` level, do it during init. This will work easily as long as we dont support dynamic config values.

paweldomas: Yes

PR #719, opened 2019-12-15
fix(zoom): pass JWT to integrations service 

virtuacoplenny:
So only validated Spot instances can request
a signed Zoom meeting number. The getJwtDomains
selector has been modified so that Zoom is
automatically injected as a JWT domain if
Zoom is enabled--this follows the pattern
set by getAllAllowedMeetingDomains.

Built on top of https://github.com/jitsi/jitsi-meet-spot/pull/718.

### DIFF
--- a/spot-client/src/common/app-state/config/selectors.js
+++ b/spot-client/src/common/app-state/config/selectors.js
@@ -127,9 +127,32 @@ export function getJoinCodeRefreshRate(state) {
  * @param {Object} state - The Redux state.
  * @returns {Array<string>}
  */
-export function getJwtDomains(state) {
+function _getConfiguredJwtDomains(state) {
     return state.config.SPOT_SERVICES.jwtDomains;
 }
+
+/**
+ * A selector which returns domains for which the backend-provided JWT is valid.
+ * Injects Zoom support if Zoom support is enabled.
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {Array<string>}
+ */
+export const getJwtDomains = createSelector(
+    _getConfiguredJwtDomains,
+    isZoomEnabled,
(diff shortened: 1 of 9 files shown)


spot-client/src/common/app-state/config/selectors.js
```diff
@@ -141,8 +166,34 @@ export function getLoggingEndpoint(state) {
...
+ * whitelist but other enabled features.
+ *
+ * @param {Object} state - The Redux state.
+ * @returns {string[]}
+ */
+export const getAllAllowedMeetingDomains = createSelector(
+    getMeetingDomainsWhitelist,
+    _getConfiguredJwtDomains,
+    isZoomEnabled,
+    (whitelistedDomains, jwtDomains, allowZoomDomains) => {
```
paweldomas: I'm wondering why jwt domains and whitelisted domains are separate config items? do we have any use case for that?

paweldomas: I see `jwtDomains` are only used in the calendar and this list merged together with the whitelisted domains, so it seems to could be just one config? Am I missing something? Are JWT domains used for anything else?

virtuacoplenny: Whitelisted domains includes all domains we wanted to be displayed in the calendar. JWT domains represents which of those domains we want to pass the JWT into the jitsi frame. It could be that the JWT is valid for Env A but spot is configured to join both Env A and Env B. If the JWT is passed into Env B there is an error and the meeting is not joined.

paweldomas: Got it

paweldomas: we need them separate then

PR #720, opened 2019-12-16
fix(error-boundary): remove View component usage

virtuacoplenny:
The View component uses RCS to notify remotes
of the TV's current view. However, if an error
occurred in RCS, it is possible that trying to
use View will cause another error in RCS,
preventing the error boundary from displaying.
Instead, just show the status overlay and
let the TV reconnect.

### DIFF
--- a/spot-client/src/common/ui/views/fatal-error.js
+++ b/spot-client/src/common/ui/views/fatal-error.js
@@ -8,8 +8,6 @@ import { ROUTES } from 'common/routing';
 
 import { Countdown, StatusOverlay } from './../components';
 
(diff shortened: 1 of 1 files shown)


paweldomas:
What the view was giving us was that it logged that the error view has been displayed. Maybe worth adding a log message that the error boundary has been displayed?

virtuacoplenny:
That is true. We could assume as much with the log `error boundary triggered` though. Should we still add an explicit log, maybe in componentDidMount?

paweldomas:
> error boundary triggered

Yeah actually that is enough.

PR #724, opened 2019-12-18
ref(zoom): inject zoom into whitelisted domain at bootstrap

virtuacoplenny:


spot-client/src/spot-tv/ui/loaders/withCalendar.js
```diff
@@ -110,10 +109,7 @@ export class CalendarLoader extends AbstractLoader {
     _loadService() {
         calendarService.setConfig(
             this.props.calendarConfig,
-            [
-                ...this.props.meetingsDomainsWhitelist,
-                ...this.props.jwtDomains
```
paweldomas: it is not clear to my why 'jwtDomains' merge goes away? I think it doesn't hurt to keep it

virtuacoplenny: jwt domains gets merged into whitelisted domains while setting default config.

paweldomas: oh I've missed this part, nice

PR #726, opened 2019-12-18
fix(zoom): pass JWT to integrations service

virtuacoplenny:


spot-client/src/spot-tv/ui/views/meeting.js
```diff
@@ -82,9 +83,9 @@ export class Meeting extends React.Component {
         this._useJwt = false;
 
         if (this._queryParams.location && props.jwt) {
-            const { host } = new URL(this._queryParams.location);
-
-            this._useJwt = props.jwtDomains.includes(host);
+            this._useJwt = this._useJwt = Boolean(findWhitelistedMeetingUrl(
```
paweldomas: what is the meaning of `this._useJwt = this._useJwt` ?

virtuacoplenny: bad copy/paste

PR #727, opened 2019-12-18
feat(webdriver): video reporter

virtuacoplenny:


virtuacoplenny:
Looks like the package doesn't have multi-remote support (multiple browsers per test) and there's a PR in the library adding support but it seems abandoned

PR #730, opened 2019-12-31
fix(deps): use jitsi fork of quietjs

virtuacoplenny:
Brief history: the queitjs fork currently in
use was supposed to be temporary, but was worked
on with the maintainer of quietjs. Another spot
team member was supposed to work on the "official"
quietjs rewrite but did not deliver so the
temporary fork continued to be used. So for
now, with ultrasound work not close to being
on the radar, the temporary fork has been combined
into jitsi's official quietjs fork.

### DIFF
--- a/spot-client/package-lock.json
+++ b/spot-client/package-lock.json
@@ -9400,8 +9400,8 @@
             }
         },
         "lib-quiet-js": {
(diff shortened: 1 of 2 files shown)


virtuacoplenny:
Force pushed to resolve a conflict in package.json.

PR #739, opened 2020-01-08
api endpoint to set muted state externally

zbettenbuk:


paweldomas:
When this external mute happens, what happens in the JitsiMeet iframe? Should JitsiMeet be aware of the muted state update as well?

PR #740, opened 2020-01-08
electron: add audio-input and allow-jit

paweldomas:
Try adding 'audio-input' because com.apple.security.device.microphone does not always work.

I don't understand the 'allow-jit' flag, but the internet says it's javascript core related, so better to have it.


saghul:
> I don't understand the 'allow-jit' flag, but the internet says it's javascript core related, so better to have it.

V8 uses a JIT which means it has memory regions markd as both writable and executable. This needs to be allowed in the sandbox.

PR #752, opened 2020-01-17
fix(xmpp): larger jitter on reconnect

virtuacoplenny:


spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -579,10 +585,14 @@ export default class XmppConnection extends Emitter {
                 return;
             }
 
+            const reconnectJitter = getJitterDelay(Math.min(4, this._reconnectsAttempted), 1000, 3);
```
paweldomas: I think `this._reconnectsAttempted += 1;` should be just above this line to start from 1

spot-client/src/common/remote-control/xmpp-connection.js
```diff
@@ -606,19 +616,16 @@ export default class XmppConnection extends Emitter {
...
-                    if (canceled) {
-                        this._isReconnecting = false;
-                        this._silentReconnectPromise = undefined;
-                    } else {
-                        setTimeout(() => {
-                            this._isReconnecting = false;
-                            this._silentReconnectPromise = undefined;
-                            canceled || this._onDisconnect(reconnectError);
-                        }, getJitterDelay(0, 5000));
+                    this._reconnectsAttempted += 1;
```
paweldomas: not here

paweldomas: LGTM otherwise

PR #754, opened 2020-01-21
fix(controller): unmount webview on app background

virtuacoplenny:


virtuacoplenny:
Jenkins test this please.

PR #756, opened 2020-01-22
Reorg native controller calls

zbettenbuk:
this is just a reorg for the native controller calls, as they were all around the app


virtuacoplenny:
Jenkins test this please.

PR #759, opened 2020-01-27
Backendless flow for fixed code segment

zbettenbuk:


virtuacoplenny:
You should try testing in an incognito to reproduce the test failures. Basically it's a JS error due to changes in generating the random join key.

Also, why is this feature desired for open source flow? And wouldn't setting of a key be better off within the admin tools of the backend flow?

zbettenbuk:
> You should try testing in an incognito to reproduce the test failures. Basically it's a JS error due to changes in generating the random join key.
> 
> Also, why is this feature desired for open source flow? And wouldn't setting of a key be better off within the admin tools of the backend flow?

Yep, for backend-enabled flows it should be on the backend, but I wanted to have it in the OS flow too, because for me the feature makes sense and the OS flow is the easier to test (and to validate the feature).

PR #764, opened 2020-02-03
Poly setup flow

zbettenbuk:


spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
```diff
@@ -77,28 +96,56 @@ export class SyncWithBackend extends React.Component {
...
+
+    /**
+     * Tries to perform an auto sync if a pairing code is provided on start.
+     *
+     * @returns {void}
+     */
+    _tryAutoSync() {
+        const { llpc } = this.props.startParams || {};
+
+        if (!llpc || llpc.length !== EXPECTED_LLPC_LENGTH) {
```
virtuacoplenny: What about validating when the params get put into redux? That way validation occurs as close to time of retrieval as possible.

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
```diff
@@ -34,12 +38,22 @@ export class SyncWithBackend extends React.Component {
         super(props);
 
         this.state = {
+            autoSyncing: false,
```
virtuacoplenny: Shouldn't this default to true? It seems like it defaults to false and just gets set to true immediately after mount. What if its default is based on if there is an "llpc"?

spot-client/src/common/app-state/setup/reducer.js
```diff
@@ -65,6 +68,12 @@ const setup = (state = DEFAULT_STATE, action) => {
             preferredSpeaker: action.speakerLabel
         };
 
+    case SET_START_PARAMS:
```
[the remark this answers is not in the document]
zbettenbuk: Actually the fact that these are called params and that query params are also called params is just a coincidence for me. These are similarly params as they were command line params, or anything else. But do you have an example how you'd call them?

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
```diff
@@ -3,12 +3,14 @@ import React from 'react';
...
 import { connect } from 'react-redux';
 
-import { addNotification, getProductName } from 'common/app-state';
+import { addNotification, getProductName, getStartParams } from 'common/app-state';
 import { CodeInput, LoadingIcon } from 'common/ui';
 import { logger } from 'common/logger';
 
 import { pairWithBackend } from '../../../app-state';
 
+const EXPECTED_LLPC_LENGTH = 8;
```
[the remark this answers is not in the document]
zbettenbuk: I got the LLPC abbreviation from the backend (long lived pairing code). Actually it quite bothers me that there are multiple types of pairing codes and all of them are just usually called "pairing code". For me this seems more like a setup code. But I guess it's too late to change it.

Thanks for the info about configurability, I take a look!

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
zbettenbuk: Actually I intentionally stayed away from doing that. As you see I preserve the query params at the start of the app in a generic way, there can be literally any keys with any values, and they are stored in the setup Redux tree. I want the features to use them in their own way: every features use what they need, and others don't need to know about them.

Now, if I want to validate these at store time, I need to put feature logic (I know this term doesn't really exist in Spot, but I guess you understand it) into the common reducer which feels kinda weird.

Or did you have something different in your mind? What do you think?

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
zbettenbuk: Yeah, I was thinking about it defaulting to true. Let me revisit this and see if I can improve.

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
virtuacoplenny: What about not validating at all then? And if you're preserving params for a use case which may not exist yet, why do you store them in setup?

spot-client/src/app.js
```diff
@@ -287,6 +290,23 @@ export class App extends React.Component {
...
+     */
+    _preserveQueryParams() {
+        const params = new URLSearchParams(this.props.location.search);
+        const rawParamsObject = {};
+
+        for (const [ key, value ] of params.entries()) {
+            rawParamsObject[key] = value;
+        }
+
+        this.props.dispatch(setStartParams(rawParamsObject));
```
virtuacoplenny: What do you think about shoving this stuff into the already cluttered bootstrap.js? Maybe where the default state is produced?

zbettenbuk: Good idea, done! Although we may need to clean up that mess there in the future

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
zbettenbuk: Done, thanks!

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
zbettenbuk: Not sure where else to store, without being feature (screen) dependant. I get the not-yet-used argument, but to be honest, it would take more time to persist them on a need basis than to persist them all and let the features (screens) use them as they want.
Also, I'm certain that we'll need more params in the future than these 3 (1 is in this commit, 2 more to come into the PR), especially if we implement further integrations.

spot-client/src/spot-tv/ui/components/setup/SyncWithBackend.js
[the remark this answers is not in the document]
zbettenbuk: Did the props stuff. Although this screen wasn't configurable. But I guess you meant to make it so.

spot-client/src/index.js
```diff
@@ -143,6 +144,17 @@ history.listen(location => {
...
+/**
+ * Persisting query params.
+ */
+const rawParamsObject = {};
+
+for (const [ key, value ] of queryParams.entries()) {
+    rawParamsObject[key] = value;
+}
+
+store.dispatch(setStartParams(rawParamsObject));
```
virtuacoplenny: You could also do this in the `createStore` call so you don't need the action, unless it will happen soon where the start params could be changed after bootstrap.

zbettenbuk: Ok, I do a follow up commit

PR #766, opened 2020-02-05
relocate start param persistence

zbettenbuk:
This is just a refactor to relocate the param persistence to a better place

PR #767, opened 2020-02-06
Param to enable volume control

zbettenbuk:
This PR allows the spot tv runtime environment to append a `volumeControlSupported` param to enable the hardware volume control functionality if the runtime environment supports that.

Please note it's rebased on top of https://github.com/jitsi/jitsi-meet-spot/pull/766 so please approve that first

### DIFF
--- a/spot-client/src/common/app-state/spot-tv/reducer.js
+++ b/spot-client/src/common/app-state/spot-tv/reducer.js
@@ -20,6 +20,7 @@ const DEFAULT_STATE = {
     tileView: false,
     videoMuted: true,
     view: undefined,
+    volumeControlSupported: false,
     waitingForMeetingStart: false
 
     // FIXME: setting the default prevents wired screensharing from being
--- a/spot-client/src/index.js
+++ b/spot-client/src/index.js
@@ -30,6 +30,7 @@ import {
     remoteControlClient
 } from 'common/remote-control';
 import {
+    getBoolean,
     getPersistedState,
     loadScript,
     setPersistedState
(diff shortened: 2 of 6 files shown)


spot-client/src/common/app-state/spot-tv/reducer.js
```diff
@@ -20,6 +20,7 @@ const DEFAULT_STATE = {
     tileView: false,
     videoMuted: true,
     view: undefined,
+    volumeControlSupported: undefined,
```
virtuacoplenny: false?

spot-client/src/index.js
```diff
@@ -58,7 +59,8 @@ const store = createStore(
         spotTv: {
 
             // Will get overridden on Spot-Remote by Spot-TV updates.
-            electron: isElectron()
+            electron: isElectron(),
+            volumeControlSupported: getBoolean(startParams.volumeControlSupported)
```
virtuacoplenny: what happens on page reload? does this need to persist?

spot-client/src/common/app-state/spot-tv/reducer.js
zbettenbuk: Well, I thought undefined means it wasn't defined explicitly, but yeah, false makes sense

spot-client/src/index.js
zbettenbuk: Good point, added. Also, implemented in a way that the URL param will always overwrite the persisted value, if present.

PR #768, opened 2020-02-06
feat(electron): add network offline handling

virtuacoplenny:
Screen Shot 2020-02-06 at 8 17 15 AM

### DIFF
--- a/spot-electron/src/application-window/applicationwindow.js
+++ b/spot-electron/src/application-window/applicationwindow.js
@@ -43,7 +43,7 @@ function createApplicationWindow() {
         clearTimeout(showCrashPageTimeout);
 
         if (isOnline) {
-            loadDefaultUrl(applicationWindow);
+            applicationWindow.loadURL(defaultSpotURL);
         } else {
             applicationWindow.loadFile('src/static/offline.html');
         }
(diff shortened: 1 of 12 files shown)


spot-electron/src/application-window/applicationwindow.js
```diff
@@ -35,26 +39,55 @@ function createApplicationWindow() {
...
+/**
+ * Loads the root Spot-TV url on the provided window.
+ *
+ * @param {BrowserWindow} browserWindow - The BrowserWindow instance in which
+ * the Spot-TV url should be opened.
+ * @returns {void}
+ */
+function loadDefaultUrl(browserWindow) {
+    browserWindow.loadURL(defaultSpotURL, {
+        userAgent: `${browserWindow.webContents.getUserAgent()} ${SPOT_ELECTRON_FEATURE_VERSION}`
```
paweldomas: I think this may be adding the spot electron version on each loadURL. I remember this because I've tried to make a method like this in the past. Maybe worth check if that's not the case?

virtuacoplenny: You were right that it was appending the version multiple times. I made a change to set the user agent outside of url load.

PR #773, opened 2020-02-10
disable auto join for wired screensharing and add a config flag

paweldomas:


spot-client/src/common/app-state/config/default-config.js
```diff
@@ -86,6 +86,12 @@ export default {
      */
     DEFAULT_MEETING_DOMAIN: process.env.DEFAULT_MEETING_DOMAIN || 'meet.jit.si',
 
+    /**
+     * When set to true the TV will automatically join a random meeting if a change to video image detected on
+     * the screensharing device occurs. In other words joins a meeting if the screensharing device gets connected.
+     */
+    ENABLE_AUTO_SS_JOIN: process.env.ENABLE_AUTO_SS_JOIN || false,
```
virtuacoplenny: Maybe this should be in temporary flags?

PR #777, opened 2020-02-12
ref(ultrasound): move initialization to static, add unit tests

virtuacoplenny:


spot-client/src/spot-remote/ultrasound/UltrasoundTransmitter.js
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Initializes a new {@code ultrasoundTransmitter} instance.
+     *
+     * @param {number} transmissionDelay - The gap between emitting ultrasonic
+     * messages.
+     * @param {string} initialText - Any text to start transmitting immediately
+     * after instantiation.
+     */
+    constructor(transmissionDelay = 3000, initialText) {
```
bbaldino: No mention of units for `transmissionDelay` (ideally in the variable name, but at least in the docs?)

spot-client/src/spot-remote/ultrasound/UltrasoundTransmitter.js
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /**
+     * Updates the text to be transmitted and initializes a {@code Transmitter}
+     * to send the text if not already initialized.
+     *
+     * @param {string} text - The message to be transmitted via ultrasound.
+     * @returns {void}
+     */
+    setMessage(text = '') {
```
bbaldino: Maybe this is due to the nature of the underlying library, but this API seems weird to me.  As a user, I'd expect to be able to call something like `sendMessage` and I'd expect this lib to hide away the details around doing that (retransmitting, etc.).

virtuacoplenny: Is that not what happens now. UltrasoundTransmitter hides away the retransmitting.

bbaldino: `setMessage` is confusing to me as a user, though.  I want to _send_ a message.  I'd think I need to call something else after `setMessage` to actually send it.  Underneath, you're setting a message and then something is checking it, but that's just confusing as a user I think.  As far as the user is concerned this is sending, so just call it `sendMessage`.

spot-client/src/spot-remote/ultrasound/UltrasoundTransmitter.js
```diff
@@ -0,0 +1,128 @@
...
+     * @param {string} text - The message to be transmitted via ultrasound.
+     * @returns {void}
+     */
+    sendMessage(text = '') {
+        const newText = (text || '').trim();
+        const oldText = this._text;
+
+        this._text = newText ? str2ab(newText) : null;
+
+        if (!this._text) {
```
bbaldino: What's the use case where an empty string would be passed to cancel the next transmission?  What about just a `stop` or `stopSending` method instead?

PR #788, opened 2020-03-11
feat: don't activate beacons on permanently paired remotes

zbettenbuk:
This PR makes the controller smarter by only activating beacon detection when the remote is waiting for a join code to be entered (not permanently paired, not in a meeting..etc).


zbettenbuk:
@paweldomas the backend tests fail with random errors. Do you think we should merge it this way, or re-run the tests until they pass?

paweldomas:
have you tried looking at the test results? there are screenshots included

paweldomas:
at glance they don't look random to me

zbettenbuk:
Well, in the last run these failed:

image

in the prev run, these failed:

image

I think this is quite random :)

paweldomas:
yeah, it might be related to the new websocket load balancing which randomly fails with 5xx

PR #790, opened 2020-03-12
fix: timing issue when restarting beacons

zbettenbuk:
The stop function on windows seems to be sync, but the status is not updated immediately, so we need to handle it differently

### DIFF
--- a/spot-electron/src/bt-beacon/daemon/win32/bluetoothd.js
+++ b/spot-electron/src/bt-beacon/daemon/win32/bluetoothd.js
@@ -30,6 +30,10 @@ class BluetoothD {
 
         this.advertiser.on('StatusChanged', ({ status }) => {
             this.status = status;
+            if (this.advertisingStatusUpdatedCallback === 'function') {
+                this.advertisingStatusUpdatedCallback(status === 2);
+                this.advertisingStatusUpdatedCallback = undefined;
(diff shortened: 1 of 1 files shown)


paweldomas:
Sorry, but I don't get it how it works or what's the fixed flow exactly from the description. Also I think  commit's message is a better place for the description rather than the PR as it will get lost here.

spot-electron/src/bt-beacon/daemon/win32/bluetoothd.js
```diff
@@ -30,6 +30,10 @@ class BluetoothD {
 
         this.advertiser.on('StatusChanged', ({ status }) => {
             this.status = status;
+            if (this.advertisingStatusUpdatedCallback === 'function') {
+                this.advertisingStatusUpdatedCallback(status === 2);
```
paweldomas: this should probably be `status !== 2`?

zbettenbuk: No, true equals running state, false equals stopped (or failed of any sort). See line 84, 85 below.

paweldomas: I still think this code is too hard to understand, but I give up

PR #792, opened 2020-03-12
ref: add additional beacon logging

zbettenbuk:
This adds a new log into a critical part of the code


zbettenbuk:
I'm merging this per our conversation as the additional logging shouldn't be a reason for the failures.

PR #799, opened 2020-03-19
feat: use spot mobile SDK in controller

zbettenbuk:
This PR moves the spot mobile SDK (from https://github.com/jitsi/jitsi-spot-sdk) into this monorepo and also integrates it into the `spot-controller` to fix a bug that didn't exist in the SDK but existed in the old repo.

When reviewing, feel free to ignore the `spot-mobile-sdk` folder as that's just a copy of the original repo with some `package.json` changes.


spot-controller/ios/spot-controller/Info.plist
```diff
@@ -38,15 +38,15 @@
 		</dict>
 	</dict>
 	<key>NSBluetoothPeripheralUsageDescription</key>
-	<string>Detect nearby meeting devices</string>
+	<string>Detect nearby Spot devices</string>
```
paweldomas: is this string updated by branding if there is any? if not it needs  to stay as it was

zbettenbuk: Yep, I wrote them back, thanks

PR #931, opened 2022-10-03
fix(build) - remove bt beacon windows

dudumanbogdan:


spot-electron/src/bt-beacon/btbeacon.js
```diff
@@ -17,8 +19,9 @@ class BTBeacon {
      */
     constructor() {
         this.region = config.getValue('beacon.region');
+        const isWin = platform === 'win32';
```
saghul, line 22: Instead check explicitly if the platform is darwin, since this is also not supported on Linux.

saghul: Is the check necessary though? Isn't `daemon` undefined?

dudumanbogdan: Default was {}, changed to export undefined and removed this logic from here

PR #940, opened 2022-11-24
feat(webdriverio) - add webdriverio changes to have tests working and running

dudumanbogdan:
- included changes to run webdriverio locally
- switched all code to `async mode` as `sync mode` will not be supported anymore starting with nodejs v16 (https://webdriver.io/docs/sync-vs-async)
- changed to chromedriver service as selenium-standalone is out of scope, repository was archived (https://github.com/webdriverio-boneyard/wdio-selenium-standalone-service)
- changed selectors from `Name Attribute` to `id/class` approach as `This selector strategy it deprecated and only works in old browser that are run by the JSONWireProtocol protocol or by using Appium.` (https://webdriver.io/docs/selectors name attribute)

Note: with chrome 106+, there is a bug introduced and async/await does not work anymore as tests are stopping on element exists / element to be displayed and does not continue. An workaround for this is to use chromedriver 105. If on the system there is a chrome with a different version / version is updated automatically by the company rules `/Applications/Google Chrome.app`, set in `wdio.config.js` - `services -> chromedriver -> chromedriverCustomPath` with the path of chromedriver 105
- setting `chromedriverCustomPath` to a path folder from inside the project (chrome driver binary 105), does not work. Working only if was outside of the project


dudumanbogdan:
Screenshot 2022-11-24 at 9 33 49 AM

saghul (approved):

PR #972, opened 2023-01-13
feat(meeting) - disable kick remotes on meeting end

dudumanbogdan:
- disable kick remotes on meeting end
- include configuration value for controlling kick remotes on meetings end


saghul (approved):
❤️

PR #973, opened 2023-01-13
fix(log) - fix parsing log when prop text has not json structure

dudumanbogdan:
- on pilot, `text` is not always a JSON
eg: event.text: "2023-01-13T07:53:14.497Z waitingForCall view updating {"activeTab":"input"}"


dudumanbogdan:
Screenshot 2023-01-13 at 9 51 15 AM
-3a0261365db8.png)

saghul (approved):

PR #975, opened 2023-01-13
fix(ui,share-button) - fix share button ui

dudumanbogdan:
- fix share button UI when screensharing is on
- issue introduced from e2e webdriverio changes: 'stop-share' is also used in another place as a container and has flex:1 which increases width of item


dudumanbogdan:
Screenshot 2023-01-13 at 12 04 07 PM

saghul (approved):

PR #976, opened 2023-01-17
feat(package,material-ui) - migrate material ui to latest

dudumanbogdan:
- update package.json and imports from @mui. Migrate to react 17 as peer-dependency is >=17 (https://mui.com/material-ui/getting-started/installation/)
- update unit tests after migrating to react 17
- remove default options using @mui codemons
- include small breaking changes after upgrade (default button has text-transform: upper)
Ex: before button text was wrapped in span thus `text-transform` was not applied. Now, text is directly in `<button>`


saghul (dismissed):
Does it all look the same? If so, let's go!

dudumanbogdan:
> Does it all look the same? If so, let's go!

Yes, it's looking the same. We didn't had big impact because seems we are using the default UI components (we don't have overrides around material ui components and logic).

I did another pushed and remove at all @emotion package as it seems we don't need it at all. Everything works :D

saghul (approved):

saghul:
Approved!

PR #1011, opened 2023-07-11
fix(client,calendars) filter events without a valid URL

saghul:


saghul:
It is a testament of how good those before me were that tests caught this is intended behavior :-)

PR #1030, opened 2023-08-23
feat(logger) send logs to aws

mihhu:


spot-electron/src/logger/utils.js
```diff
@@ -0,0 +1,30 @@
...
+    /**
+     * Builds the log message.
+    */
+    parseLogMessage: (level, message) => `[${LOG_LEVEL[level]}] ${message}`,
+
+    /**
+     * Returns the host identifier.
+     */
+    getHostId: () => {
+        const ipAddress = [].concat(...Object.values(os.networkInterfaces()))
```
saghul, line 25: Why do we want to have the IP address here?

spot-electron/main.js
```diff
@@ -28,6 +29,12 @@ require('./src/client-control');
 require('./src/exit');
 require('./src/volume-control');
 
+process.env.USE_CLOUDWATCH_LOGS && awsLogger.maybeCreateLogStream()
```
saghul, line 32: How is the Electron app going to get these env variables?

saghul (commented):
1 minor comment, but also a larger one: how is the app going to know about the logging details?

spot-electron/src/logger/utils.js
mihhu: In the case of cloudwatch, I use it as the log stream name, and for s3 I prefix the logs with it. It's all in order to identify which machine was running the meeting. Maybe there's a better way to achieve this, please advise if so.

spot-electron/main.js
mihhu: I was thinking that the person that installs the app can set up these too.

PR #1031, opened 2023-08-24
feat(ui) move room info from tv to remote

mihhu:


spot-client/src/spot-tv/ui/views/spot-view.js
```diff
@@ -4,9 +4,6 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import { connect } from 'react-redux';
 
-
-import { JoinInfo } from './../components';
```
saghul, line 8: Is this component used elsewhere? If not, please kill it.

saghul (commented):
Other than the meeting name we need to display the sharing domain and code.

Can you add a screenshot too please?

spot-client/src/spot-tv/ui/views/spot-view.js
mihhu: It's still used on the [/tv page](https://github.com/jitsi/jitsi-meet-spot/blob/master/spot-client/src/spot-tv/ui/views/home.js#L124)

mihhu:
image

How about this?

saghul:
> image
> 
> How about this?

I'd drop the name of the room from under the meeting name, since it's already at the top. LGTM otherwise!

saghul (approved):

PR #1032, opened 2023-08-25
feat(setup) select preferred resolution

mihhu:


spot-client/src/spot-tv/ui/components/setup/select-media/select-media.js
```diff
@@ -235,6 +252,13 @@ class SelectMedia extends React.Component {
     _getDefaultDeviceListState() {
         return {
             cameras: [],
+            resolutions: [
+                {
+                    label: this.props.t('setup.noSelection'),
+                    value: ''
+                },
+                ...RESOLUTIONS
```
saghul, line 260: Please make the labels translatable.

spot-client/src/common/utils/meeting.js
```diff
@@ -129,3 +145,34 @@ export function parseMeetingUrl(url) {
...
+ * @param {string} preferredResolution - The resolution of choice.
+ * @returns {boolean}
+ */
+export const getVideoConstraints = (preferredAspectRatio, preferredResolution) => {
+    const aspectRatio = Number(preferredAspectRatio);
+    const resolution = Number(preferredResolution);
+
+    return {
+        video: {
+            height: {
```
saghul, line 162: You need to add a `aspectRatio: 16 / 9` constraint or 4K won't work.

spot-client/src/common/utils/meeting.js
```diff
@@ -129,3 +145,34 @@ export function parseMeetingUrl(url) {
...
 }
+
+/**
+ * Builds the video constraints based on the preferred aspect ratio and resolution.
+ *
+ * @param {string} preferredAspectRatio - The aspect ratio of choice.
+ * @param {string} preferredResolution - The resolution of choice.
+ * @returns {boolean}
+ */
+export const getVideoConstraints = (preferredAspectRatio, preferredResolution) => {
```
saghul, line 156: Where does the aspect ratio come from? I think we should use 16 / 9 at all times. Regardless of the size of the app window, cameras support stabndard aspect ratios, and requiring 16 / 9 for an in-room device sounds very reasonable.

saghul (commented):
Left some comments, PTAL!

spot-client/src/common/utils/meeting.js
mihhu: I used [this](https://github.com/jitsi/jitsi-meet-spot/blob/master/spot-client/src/common/app-state/config/selectors.js#L276), but I'll switch it to 16 / 9.

PR #1035, opened 2023-08-29
fix(dial-pad) resize dial buttons based on screen height

mihhu:


saghul (approved):
LGTM, but can you pl share a screenshot before merging?

mihhu:
> LGTM, but can you pl share a screenshot before merging?

image
image
image

The last one would be the latest iPad Pro max resolution, should I maybe resize the buttons even more for that scenario? Please let me know if you want to see screenshots for other dimensions.

saghul:
I think this is a great improvement already!

PR #1039, opened 2023-09-15
feat(spot-client) refactor conflict detection

saghul:
Instead of showing a dedicated page with a retry option just keep retrying in the background.

An error notification is shown to indicate the problem.


mihhu (approved):

saghul:
Screen Shot 2023-09-15 at 13 09 34

I managed to test it!

PR #1054, opened 2023-11-21
feat(spot-client) drop API health-check

saghul:
It's prone to false positives. The meeting itself can generate an error and the user can hangup from the remote if there is a problem. Automating this is dangerous because we don't know what can happen that may cause a false positive.


paweldomas:
Note that the health check is not communicating with a remote backend service of any kind, but rather sends command to the embedded jitsi iframe. If I recall, that command was a plain get operation without any async access. If it's stuck, it is probably something weird going on there and I'd recommend investigating that.

Now if the iframe stops being responsive, what is likely to happen is the user won't be able to control it anymore. I would recommend as the first step putting it as an error log and maybe report an analytics event to see when and how often it happens. It may reveal some problems.

saghul:
Yes, I noticed what the test does, but it's weird it the meeting was working until it timed out. Granted 5 seconds _should_ be enough for it to work, but since somwe of these systems are used in very sensitive environments, not failing hard has become a priority. A lockup can be recovered from, but a hard failure not.

saghul:
I found some other weird shit like the `onload` event not being triggered, WTF. I'll revisit this if need be.

PR #1080, opened 2024-08-05
feat(client): improve modal and add hang up confirmation

andrei-gavrilescu:
Add some enhancements to modal (make it can't be clicked through)
Screenshot 2024-08-05 at 11 22 47
Add hang up confirmation.
Screenshot 2024-08-05 at 11 53 53

### DIFF
--- a/spot-client/src/common/css/_hangup-modal.scss
+++ b/spot-client/src/common/css/_hangup-modal.scss
@@ -1,0 +1,31 @@
+.hangup-modal-select {
+    @include centered-content;
+    flex-direction: column;
+    justify-content: center;
+    min-width: 360px;
+    text-align: center;
+    width: 100%;
+
+    .content {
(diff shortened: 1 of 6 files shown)


saghul (approved):
Excellent work! Pl try to run the e2e tests locally, I bet some of them need to be adjusted for the hangup change.

There are some that fail in the CI, which we can keep ignoring for a while longer.

andrei-gavrilescu:
> Excellent work! Pl try to run the e2e tests locally, I bet some of them need to be adjusted for the hangup change.
> 
> There are some that fail in the CI, which we can keep ignoring for a while longer.

Seems like these fail randomly, it might have something to do with the duplicate events problem, I'll check to see if adding a fix there makes them more consistent.

saghul (approved):

PR #1081, opened 2024-08-14
Better random

saghul:
Sorry, got nerdsnipped and now you have you review this :-P


spot-client/src/common/utils/random.js
```diff
@@ -1,12 +1,87 @@
+/**
+ * Helper class to generate random values one by one, using an auxiliary
```
andrei-gavrilescu, line 2: Neat trick!

spot-client/src/common/utils/random.js
```diff
@@ -1,12 +1,87 @@
...
+        r.push(seq[idx]);
+    }
+
+    return r.join('');
+}
+
+/**
+ * This alphabet removes all potential ambiguous symbols, so it's well suited for a code.
+ */
+const BASE32 = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ234567';
```
andrei-gavrilescu, line 63: This is cuel. Something comes to mind though, users won't know that they can only enter non ambiguous chars, so on the remote side when they enter the code they still might confuse 0 with  O, maybe remove O and I as well.

andrei-gavrilescu (approved):

spot-client/src/common/utils/random.js
saghul: I'd like to stay with the base32 standard. A monospaced font should take care of that...

PR #1082, opened 2024-08-21
fix(client): deduplicate messages / lower case join code

andrei-gavrilescu:


spot-client/src/common/remote-control/remoteControlClient.js
```diff
@@ -594,13 +594,15 @@ export class RemoteControlClient extends BaseRemoteControlService {
...
      */
     _onSpotTvStatusReceived(from, state) {
-        // NOTE There will be no timestamp in the initial Spot TV presence
-        const { timestamp } = state;
+        const { msgId } = state;
 
-        const currentTimestamp = this._lastSpotState && this._lastSpotState.timestamp;
-
-        if (currentTimestamp && timestamp && currentTimestamp >= timestamp) {
+        const previousMsgId = this._lastSpotState && this._lastSpotState.msgId;
```
saghul, line 599: Use optional chaining?

spot-client/src/common/remote-control/remoteControlServer.js
```diff
@@ -272,7 +275,7 @@ export class RemoteControlServer extends BaseRemoteControlService {
      * @returns {Promise} - A Promise resolved when the code has been refreshed.
      */
     refreshJoinCodeWithXmpp() {
-        const roomLock = generateRandomString(3);
+        const roomLock = generateRandomString(3).toLowerCase();
```
saghul, line 278: I don't think the passcode needs to be lowercase, does it?

spot-client/src/common/remote-control/remoteControlServer.js
```diff
@@ -82,7 +85,7 @@ export class RemoteControlServer extends BaseRemoteControlService {
             // If there's no joinCode service then create a room and let the lock
             // be set later. Setting the lock on join will throw an error about
             // not being authorized..
-            roomName: generateRandomString(3)
+            roomName: generateRandomString(3).toLowerCase()
```
saghul, line 88: What about above, when we split a 6 digit code?

saghul (approved):

spot-client/src/common/remote-control/remoteControlServer.js
andrei-gavrilescu: It's not that the passcode needs to be lowercase, but when you enter it on the remote, even if the ui shows them as upper case it does a toLowerCase() and then sends it to the server.
https://github.com/jitsi/jitsi-meet-spot/blob/7f340e04c98710291338838e160b611913cd857a/spot-client/src/spot-remote/ui/views/join-code-entry.js#L310

saghul (approved):

PR #1130, opened 2025-06-16
feat(client): add recording consent dialog

andrei-gavrilescu:
Whenever the recording consent dialog pops up on jitsi-meet also show it in the spot remote.
Screenshot 2025-06-13 at 15 04 20

Depends on: https://github.com/jitsi/jitsi-meet/pull/16141

### DIFF
--- a/spot-client/package.json
+++ b/spot-client/package.json
@@ -76,8 +76,8 @@
         "build:dev": "webpack --config ./webpack.config.js --progress",
         "build:prod": "NODE_ENV=production webpack --config ./webpack.config.js --progress",
         "docs": "jsdoc -c ./jsdoc.config.js",
-        "lint": "eslint .",
-        "lint:fix": "eslint . --fix",
+        "lint": "eslint . --max-warnings=0",
+        "lint:fix": "eslint . --fix --max-warnings=0",
         "start:dev": "webpack-dev-server",
         "test": "jest"
     },
--- a/spot-client/package.json
+++ b/spot-client/package.json
@@ -77,6 +77,7 @@
         "build:prod": "NODE_ENV=production webpack --config ./webpack.config.js --progress",
         "docs": "jsdoc -c ./jsdoc.config.js",
         "lint": "eslint .",
+        "lint:fix": "eslint . --fix",
         "start:dev": "webpack-dev-server",
         "test": "jest"
     },
--- a/spot-client/src/common/css/_recording-consent-modal.scss
+++ b/spot-client/src/common/css/_recording-consent-modal.scss
@@ -1,0 +1,41 @@
+.recording-consent-modal {
+    @include centered-content;
+    flex-direction: column;
+    font-family: 'Arial', sans-serif;
+    min-width: 360px;
+    width: 600px;
+
+    .content {
+        display: flex;
+        flex-direction: column;
+    }
+
+    .footer {
+        display: flex;
+        gap: 12px;
+    }
+    
+   // Rule that applies to all classes starting with 'understand-button'
+    [class*="understand-button"] {
+        flex: 1;
+        font-size: 21px;
+        font-weight: 600;
+    }
+    
+    .understand-button-unmute {
+        background-color: white;
+        color: black;
+    }
+
+    .description {
+        font-size: 21px;
+        font-weight: 400;
+        padding-bottom: 40px;
+    }
+
+    .title {
+        font-size: 29px;
+        font-weight: 600;
+        padding-bottom: 40px;
+    }
+}
--- a/spot-client/src/spot-remote/ui/components/recording-consent/index.js
+++ b/spot-client/src/spot-remote/ui/components/recording-consent/index.js
@@ -1,0 +1,1 @@
+export { default as RecordingConsentDialog } from './RecordingConsentDialog';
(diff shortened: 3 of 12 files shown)


spot-client/package.json
```diff
@@ -77,6 +77,7 @@
         "build:prod": "NODE_ENV=production webpack --config ./webpack.config.js --progress",
         "docs": "jsdoc -c ./jsdoc.config.js",
         "lint": "eslint .",
+        "lint:fix": "eslint . --fix",
```
saghul: While you're here, add `--max-warnings=0` to both linting commands please.

spot-client/src/common/css/_recording-consent-modal.scss
```diff
@@ -0,0 +1,41 @@
+.recording-consent-modal {
```
saghul, line 1: Doesn't `modal.scss` suffice?

spot-client/src/spot-remote/ui/components/recording-consent/index.js
```diff
@@ -0,0 +1 @@
+export { default as RecordingConsentDialog } from './RecordingConsentDialog';
```
saghul: Let's drop the index and import from the file directly.

I know that's not quite the pattern used across this particular codebase, but we have moved away from vaning too many index files on JM and it has been A Good Thing (TM).

So let's start applying it here too.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
```diff
@@ -40,8 +42,10 @@ export class InCall extends React.Component {
         meetingDisplayName: PropTypes.string,
         onForceStopWirelessScreenshare: PropTypes.func,
         onHangUp: PropTypes.func,
+        onShowModal: PropTypes.func, // Add this line
```
saghul: What's with the comment?

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
```diff
@@ -40,8 +42,10 @@ export class InCall extends React.Component {
...
         onShowScreenshareModal: PropTypes.func,
         onSubmitPassword: PropTypes.func,
+        recordingConsentDialogOpen: PropTypes.bool, // Add this line
```
saghul: Ditto.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
```diff
@@ -81,11 +85,25 @@ export class InCall extends React.Component {
...
+        if (this._showCancelMeetingTimeout && inMeeting) {
             clearTimeout(this._showCancelMeetingTimeout);
             this._showCancelMeetingTimeout = null;
         }
+
+        // Only show or hide the recording consent dialog if its state has changed
+        if (!prevDialogOpen && currentDialogOpen) {
+            this.props.onShowModal(RecordingConsentDialog);
+        } else if (prevDialogOpen && !currentDialogOpen) {
+            this.props.hideModal(); // Use this.props to avoid shadowing
```
saghul: Is it necessary to manually hide the modal?

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
```diff
@@ -221,6 +241,16 @@ function mapDispatchToProps(dispatch) {
             return dispatch(hangUp(true));
         },
 
+        /**
+         * Shows a modal with the specified component.
+         *
+         * @param {React.Component} component - The component to show in the modal.
+         * @returns {void}
+         */
+        onShowModal(component) {
```
saghul, line 248: Is it the 1st time we have needed this? Are there no other similar dialogs, like the hangup confirmation? What flow do they follow?

saghul (commented):
Excellent work! Left some comments.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
andrei-gavrilescu: It's the first time we needed a modal dialog that reacts to a meeting state change, unlike the other modals that react when an action happens on the remote, like button press.
I initially tried to have the control in middleware (felt more appropriate) but because the way the remote and the tv work, a common code base that shares logic around redux but with two different states on two machines  (or more) that sync  T_T, I started running into weird situations. It was necessary to programmatically distinguish if we're on the tv or on the remote in the same middleware... things started getting weird so I decided to move the control here, were we're sure that just the remote will react.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
andrei-gavrilescu: I think so, otherwise there's no logic that will hide it. We could have it hide after the action is performed, but I wanted it to be in sync with what we see on the tv side of things.
So the flow is you click a button- -> goes to tv -> tv reacts -> state from tv comes back -> closes the dialog.

spot-client/src/spot-remote/ui/views/remote-views/in-call.js
andrei-gavrilescu: Huh, I tried to copilot reviews with direct fixes, seems to have added those. will remove.

spot-client/src/common/css/_recording-consent-modal.scss
andrei-gavrilescu: I tried to mimic the recording dialog from meet, so it has a fixed size plus a bunch of other custom stuff.

spot-client/package.json
andrei-gavrilescu: added

saghul (approved):
Great work!

